PR #14, opened 2023-02-27
Add support for Go runtime metrics

liustanley:
### What does this PR do?

Adds code in OTLP metrics translator to map OTel runtime metrics to Datadog runtime metrics. If a mapping is found for an incoming runtime metric, both the original OTel runtime metric and the equivalent Datadog runtime metric will be sent to the backend. Runtime metric mappings are implemented for Go in a separate file containing a Go struct.

### DIFF
--- a/pkg/otlp/metrics/metrics_translator.go
+++ b/pkg/otlp/metrics/metrics_translator.go
@@ -33,6 +33,22 @@ import (
 	"github.com/DataDog/opentelemetry-mapping-go/pkg/quantile"
 )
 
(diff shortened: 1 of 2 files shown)


gbbr (approved):

mx-psi (commented):
It would be good to have a changelog note on this, just copying the one from the previous PR (how to do this is mentioned [here](https://github.com/DataDog/opentelemetry-mapping-go/blob/main/CONTRIBUTING.md#pull-requests))

PR #20, opened 2023-03-13
Add more histogram and exponential histogram tests

mx-psi:
### What does this PR do?

The first few commits add some missing tests based on coverage:
   - Add min/max to tests about histograms
   - Add test with empty histogram
   - Add test with unsupported combination of options
   - Add test for Cols

The last two commits add a new `pkg/internal/sketchtest` module for defining common test helpers for tests in `pkg/quantile` and `pkg/otlp/metrics`. These helpers help define the [quantile function](https://en.wikipedia.org/wiki/Quantile_function) of different common distributions. In particular I added the (truncated) normal and exponential distributions to have more real-world looking tests (since I would expect both of these to be common in the wild). The naming of parameters is based on how they appear on Wikipedia.

I then use this new module to add more tests, notably one on `pkg/otlp/metrics` that tests the accuracy of the output `quantile.Sketch` from a given Exponential Histogram.

### Motivation

Improve our test coverage for histograms and make sure our mapping is sound.

### DIFF
--- a/pkg/internal/sketchtest/quantile.go
+++ b/pkg/internal/sketchtest/quantile.go
@@ -1,0 +1,112 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2022-present Datadog, Inc.
+
+// package distrotest is an internal module with test helpers for generating points from a given distribution.
+package sketchtest
+
+import "math"
+
+// QuantileFunction for a distribution. The function MUST be defined in (0,1).
+// The function MAY return the minimum at 0 and the maximum at 1, for distributions where this
+// makes sense.
+type QuantileFunction func(float64) float64
+
+// CDF is the cumulative distribution function of a distribution (inverse of QuantileFunction).
+type CDF func(float64) float64
+
+// UniformQ returns the quantile function for an uniform distribution
+// with parameters a and b.
+// See https://en.wikipedia.org/wiki/Continuous_uniform_distribution
+func UniformQ(a, b float64) QuantileFunction {
+	return func(q float64) float64 {
+		return (b-a)*q + a
+	}
+}
+
+// UQuadraticQ returns the quantile function for an U-Quadratic distribution
+// with parameters a and b.
+// See https://en.wikipedia.org/wiki/U-quadratic_distribution
+func UQuadraticQ(a, b float64) QuantileFunction {
+	return func(q float64) float64 {
+		alpha := 12.0 / math.Pow(b-a, 3)
+		beta := (b + a) / 2.0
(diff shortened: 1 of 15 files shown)


pkg/quantile/ddsketch_test.go
```diff
@@ -7,19 +7,20 @@ package quantile
...
 	"github.com/DataDog/sketches-go/ddsketch"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+
+	"github.com/DataDog/opentelemetry-mapping-go/pkg/internal/sketchtest"
 )
 
 const (
-	acceptableFloatError = 2e-12
+	acceptableFloatError = 2e-11
```
mx-psi, line 20: Had to bump this after adding exponential and normal tests. I think it's just a fact that the `math.Exp` and `math.Erf`/`math.InvErf` functions do not have a lot of accuracy

pkg/otlp/metrics/sketches_test.go
```diff
@@ -187,21 +195,27 @@ func TestHistogramSketches(t *testing.T) {
 	}
 }
 
-func TestExactSumCount(t *testing.T) {
+func TestExactSummaryStatistics(t *testing.T) {
```
songy23, line 203: Test name is a bit confusing as this is testing histogram rather than summary. Consider sth. like TestExactHistogramStats

pkg/internal/sketchtest/quantile.go
```diff
@@ -0,0 +1,112 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2022-present Datadog, Inc.
+
+// package distrotest is an internal module with test helpers for generating points from a given distribution.
```
songy23, line 6: distrotest -> sketchtest

songy23 (approved):
LGTM with some non-blocking comments

pkg/otlp/metrics/sketches_test.go
mx-psi: Good point, I was using the term in [this sense](https://en.wikipedia.org/wiki/Summary_statistics) but it's definitely confusing. Renamed with your suggestion

pkg/otlp/metrics/sketches_test.go
```diff
@@ -364,6 +364,17 @@ func TestInfiniteBounds(t *testing.T) {
 		name    string
 		getHist func() pmetric.Metrics
 	}{
+		{
+			name: "(-inf, inf): 0",
```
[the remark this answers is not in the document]
mx-psi: I added an `isEmpty` parameter to the tests to ensure that we do not accidentally skip tests because of malformed data

PR #21, opened 2023-03-14
[pkg/otlp/attributes] Deprecate and replace functions which have usePreviewRules parameter.

mackjmr:
### What does this PR do?
This PR marks the exported functions which use the `usePreviewRules` parameter as deprecated, and adds new replacement functions without parameter `usePreviewRules`.

This is the first step to graduate the `exporter.datadog.hostname.preview` feature gate to stable.

Once this change is released, the next steps are:
- Marking the `exporter.datadog.hostname.preview` feature gate as stable (cannot remove it completely yet -- see [issue](https://github.com/open-telemetry/opentelemetry-collector/issues/7044)) and replace deprecated functions calls by new function calls in `opentelemetry-collector-contrib` and `datadog-agent`.
- Remove deprecated functions and codepaths which use preview rules in `opentelemetry-mapping-go`.
- [optional] Rename functions to original name following approach [here](https://github.com/open-telemetry/opentelemetry-collector/blob/main/CONTRIBUTING.md#example-3---changing-the-arguments-of-a-function).

### Motivation
https://datadoghq.atlassian.net/browse/OTEL-231

### DIFF
--- a/.chloggen/mackjmr-deprecate-usePreviewRules.yaml
+++ b/.chloggen/mackjmr-deprecate-usePreviewRules.yaml
@@ -5,7 +5,7 @@ change_type: deprecation
 component: pkg/otlp/attributes
 
 # A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Deprecate and replace functions which have usePreviewRules parameter.
+note: Deprecate and replace functions which have usePreviewRules parameter in preparation to graduate the exporter.datadog.hostname.preview feature gate to stable.
 
 # The PR related to this change
 issues: [21]
--- a/.chloggen/mackjmr-deprecate-usePreviewRules.yaml
+++ b/.chloggen/mackjmr-deprecate-usePreviewRules.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: deprecation
+
+# The name of the component (e.g. pkg/quantile)
+component: pkg/otlp/attributes
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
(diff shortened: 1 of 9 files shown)


.chloggen/mackjmr-deprecate-usePreviewRules.yaml
```diff
@@ -0,0 +1,11 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: deprecation
+
+# The name of the component (e.g. pkg/quantile)
+component: pkg/otlp/attributes
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Deprecate and replace functions which have usePreviewRules parameter.
```
liustanley, line 8: I think it would be helpful to include some context here on why we're making this change (i.e. preparing to graduate the `exporter.datadog.hostname.preview` feature gate to stable)

mackjmr: Added the context 🙂

pkg/otlp/attributes/azure/azure.go
```diff
@@ -31,8 +31,9 @@ type HostInfo struct {
 	HostAliases []string
 }
 
-// HostInfoFromAttributes gets Azure host info from attributes following
-// OpenTelemetry semantic conventions
+// HostInfoFromAttributes gets Azure host info from attributes following OpenTelemetry
+// semantic conventions.
+// Deprecated: AttributeHostID will be used as the hostname once the preview rule is stable.
```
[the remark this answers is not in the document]
mackjmr: This function is a no-op when `usePreviewRules` is `true`, so I don't believe we need a replacement function.

liustanley (approved):
LGTM

PR #23, opened 2023-03-21
[pkg/otlp/metrics] Send min and max metrics for delta histograms; rename option to send aggregations

mx-psi:
### What does this PR do?

- Send min and max histogram aggregations for delta histograms and delta exponential histograms when using the `metrics.WithHistogramAggregations` option.
- Deprecate `metrics.WithSumCountMetrics` in favor of `metrics.WithHistogramAggregations`

### Motivation

Allow backend to leverage these values when available.

### DIFF
--- a/.chloggen/mx-psi_min-max-metrics1.yaml
+++ b/.chloggen/mx-psi_min-max-metrics1.yaml
@@ -11,6 +11,9 @@ note: Deprecate metrics.WithCountSumMetrics translator option.
 issues: [23]
 
 # (Optional) One or more lines of additional information to render under the primary note.
+
 # These lines will be padded with 2 spaces and then inserted directly into the document.
 # Use pipe (|) for multiline entries.
-subtext: Use metrics.WithHistogramAggregations instead.
+subtext: | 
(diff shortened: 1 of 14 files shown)


.chloggen/mx-psi_min-max-metrics1.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: deprecation 
+
+# The name of the component (e.g. pkg/quantile)
+component: pkg/otlp/metrics
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Deprecate metrics.WithCountSumMetrics translator option.
```
songy23, line 8: May be worth mentioning that `WithCountSumMetrics` will also send min and max (if present) moving forward.

songy23 (approved):
LGTM

.chloggen/mx-psi_min-max-metrics1.yaml
mx-psi: Makes sense, done!

PR #26, opened 2023-03-27
[pkg/otlp/metrics] Use min/max on cumulative OTLP Histograms when possible

mx-psi:
### What does this PR do?

A cumulative OTLP Histogram can have a minimum and maximum fields. These represent the global extrema since the timeseries started.

This PR makes use of minimum and maximum values on the conversion of cumulative OTLP Histograms to (delta) Datadog sketches in two ways:

1. When the minimum (resp. maximum) on a point is lower (resp. higher) than the one we saw before, we know it happened in the last time window and thus can use it on the resulting sketch.
2. We can use the minimum and maximum on a point to "clamp" the estimates done on the histogram-to-sketch conversion, since we know these are global bounds for all points.

### Motivation

Improve accuracy of cumulative histogram to (delta) Datadog sketch conversion.


pkg/otlp/metrics/ttlcache.go
```diff
@@ -49,6 +49,23 @@ func (t *ttlCache) MonotonicDiff(dimensions *Dimensions, startTs, ts uint64, val
 	return t.putAndGetDiff(dimensions, true, startTs, ts, val)
 }
 
+// isNotFirstPoint determines if this is NOT the first point on a cumulative series:
+// https://github.com/open-telemetry/opentelemetry-specification/blob/v1.19.0/specification/metrics/data-model.md#resets-and-gaps
+func isNotFirstPoint(startTs, ts, oldStartTs uint64) (isNotFirst bool) {
+	// This is written down as an 'if' because I feel it is easier to understand than with a boolean expression.
```
dineshg13, line 55: nit: avoid first person statements in comments/

dineshg13 (approved):
LGTM

pkg/otlp/metrics/ttlcache.go
mx-psi, line 55:
```suggestion
```

PR #27, opened 2023-03-28
Add dependabot

mx-psi:
### What does this PR do?

Adds Dependabot configuration file. Tooling and CI are updated monthly, main modules are updated weekly.

### Motivation

Keep dependencies updated.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -1,0 +1,58 @@
+version: 2
+updates:
+  - package-ecosystem: gomod
+    directory: /pkg/otlp/metrics
+    labels:
+      - dependencies
+    schedule:
+      interval: weekly
+    ignore:
+      # Ignore internal modules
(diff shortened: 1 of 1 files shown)


.github/dependabot.yaml
```diff
@@ -0,0 +1,58 @@
+version: 2
+updates:
+  - package-ecosystem: gomod
+    directory: /pkg/otlp/metrics
+    labels:
+      - dependencies
+    schedule:
+      interval: weekly
```
dineshg13, line 8: Can we manually trigger one run of the bot ?

mx-psi: Once this is merged we can trigger a run but it's not possible before merging this. For reference, it is a bit hidden, you have to navigate to Insights > Dependency graph > Dependabot > Click on "Last checked ..." > Click on "Check for updates". 

I can trigger a run after we merge this and see if there are any issues.

dineshg13 (approved):

PR #29, opened 2023-03-31
Add Makefile targets for breaking changes

mx-psi:
### What does this PR do?

Add `apidiff-generate` and `apidiff-compare` Makefile targets.

### Motivation

A follow up PR will create a CI check that generates the gcexport data for the main branch and compares it against the current branch with `apidiff-compare`.

This will help us identify breaking changes, so that we only do them in accordance to our deprecation guidelines (#1) and by bumping the version correctly.


songy23 (approved):
LGTM

PR #65, opened 2023-05-04
[pkg/otlp/attributes] Handle literal 'host' tag

mx-psi:
### What does this PR do?

Handle literal 'host' attribute when determining the `host` tag value.

### Motivation

Not doing so causes two `host` tags/attributes to be present on all signal types under certain conditions.

### DIFF
--- a/.chloggen/mx-psi_literal-host.yaml
+++ b/.chloggen/mx-psi_literal-host.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component (e.g. pkg/quantile)
+component: pkg/otlp/attributes
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Handle literal 'host' attribute to avoid double tagging.
+
+# The PR related to this change
+issues: []
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: This avoids double tagging when using `metrics::resource_attributes_as_tags` metrics, or in all cases for traces/logs.
--- a/pkg/otlp/attributes/source.go
+++ b/pkg/otlp/attributes/source.go
@@ -29,6 +29,9 @@ const (
 	AttributeDatadogHostname = "datadog.host.name"
 	// AttributeK8sNodeName the datadog k8s node name attribute
 	AttributeK8sNodeName = "k8s.node.name"
+	// Attribute host is a literal host tag.
+	// We check for this to avoid double tagging.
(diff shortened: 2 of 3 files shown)


pkg/otlp/attributes/source.go
```diff
@@ -48,15 +51,17 @@ func getClusterName(attrs pcommon.Map) (string, bool) {
...
-//  3. the Kubernetes node name (and cluster name if available),
+//  3. cloud provider specific hostname for AWS, Azure or GCP,
 //
-//  4. the cloud provider host ID and
+//  4. the Kubernetes node name (and cluster name if available),
 //
-//  5. the host.name attribute.
+//  5. the cloud provider host ID and
+//
+//  6. the host.name attribute.
```
mx-psi, line 64: This is just adding the top line and bumping the numbers on the rest of items, but the resulting diff is not very readable

pkg/otlp/attributes/source.go
```diff
@@ -29,6 +29,9 @@ const (
 	AttributeDatadogHostname = "datadog.host.name"
 	// AttributeK8sNodeName the datadog k8s node name attribute
 	AttributeK8sNodeName = "k8s.node.name"
+	// Attribute host is a literal host tag.
+	// We check for this to avoid double tagging.
+	AttributeHost = "host"
```
songy23, line 34: IMO it is very confusing that we treat `host`, `host.name` and `datadog.host.name` in subtly different order. This has led to several customer issues.

(Not suggesting to fix it in this PR, but something that needs better documentation)

songy23 (approved):

pkg/otlp/attributes/source.go
mx-psi: That's fair, I think some subtlety is required here but maybe we can do better. For this PR, I am convinced that `host` needs to go first, since otherwise you get double tagging in some cases, so I will merge, but we should discuss how to improve the situation for the rest.

.chloggen/mx-psi_literal-host.yaml
```diff
@@ -0,0 +1,16 @@
...
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Handle literal 'host' attribute to avoid double tagging.
+
+# The PR related to this change
+issues: []
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: This avoids double tagging when using `metrics::resource_attributes_as_tags` metrics, or in all cases for traces/logs.
```
gbbr, line 16: What is double tagging? Is someone reading this expected to know? The same question for all inline comments in this PR. Why is it assumed that a newcomer/reader of this PR will know what that is? What is "double tagging"? Sounds like something impossible 😓

gbbr: Never mind, it is better explained downstairs. Resolving :)

PR #81, opened 2023-05-23
Add RuntimeMetricsTelemetry return value for MapMetrics

liustanley:
### What does this PR do?

Adds new return value on MapMetrics `RuntimeMetricsTelemetry`

### Motivation

`RuntimeMetricsTelemetry` will be used by the Agent and the Collector when determining whether to report telemetry metrics (i.e. `datadog.agent.otlp.runtime_metrics`)


pkg/otlp/metrics/metrics_translator.go
```diff
@@ -575,6 +577,7 @@ func mapHistogramRuntimeMetricWithAttributes(md pmetric.Metric, metricsArray pme
 
 // MapMetrics maps OTLP metrics into the DataDog format
 func (t *Translator) MapMetrics(ctx context.Context, md pmetric.Metrics, consumer Consumer) error {
+	t.hasRuntimeMetrics = false
```
songy23, line 580: Where else is it used? If it needs to be reset on every call to `MapMetrics`, why not just make it a return value>

liustanley: As far as I know, only once in the Agent and once in the Collector, but I wasn't sure if it's used anywhere else.

songy23 (approved):
LGTM

pkg/otlp/metrics/metrics_translator.go
```diff
@@ -482,6 +489,16 @@ func (t *Translator) source(m pcommon.Map) (source.Source, error) {
 	return src, nil
 }
 
+// extractLanguageTag edits the given languageTags slice if a new language tag is found from the given name
```
songy23, line 492:
```suggestion
// extractLanguageTag appends a new language tag to  languageTags if a new language tag is found from the given name
```

pkg/otlp/metrics/metrics_translator.go
```diff
@@ -482,6 +489,16 @@ func (t *Translator) source(m pcommon.Map) (source.Source, error) {
...
+func extractLanguageTag(name string, languageTags []string) []string {
+	for prefix, lang := range runtimeMetricPrefixLanguageMap {
+		if strings.HasPrefix(name, prefix) && !slices.Contains(languageTags, lang) {
```
songy23, line 495:
```suggestion
		if !slices.Contains(languageTags, lang) 
 && strings.HasPrefix(name, prefix) {
```

songy23 (approved):

pkg/otlp/metrics/metrics_translator.go
```diff
@@ -54,6 +55,12 @@ type Translator struct {
 	cfg     translatorConfig
 }
 
+// RuntimeMetricsTelemetry provides runtime metrics information on a MapMetrics run
```
gbbr, line 58: What type of information does this provide? How is it used? It is a struct with no exported fields and no methods attached. No information is available to the caller of MapMetrics.

pkg/otlp/metrics/metrics_translator.go
```diff
@@ -54,6 +55,12 @@ type Translator struct {
...
+type RuntimeMetricsTelemetry struct {
```
gbbr, line 59: I would rename this to `Metadata` and document it as such:
```go
// Metadata specifies information about the outcome of the MapMetrics call.
type Metadata struct {
    // RuntimeMetrics reports whether runtime metrics where found during metrics mapping.
    RuntimeMetrics bool
    // Languages specifies a list of languages for which runtime metrics where found.
    Languages []string
}
```
Lastly -- why do you need `HasRuntimeMetrics`. Isn't doing `len(Languages) > 0` the same thing? If yes, please remove the boolean. And if yes, you may as well just return a set of tags. But keeping the struct is also a good option because I suspect we may want to return more metadata eventually.

pkg/otlp/metrics/metrics_translator.go
gbbr: I see you fixed that in a follow-up PR. So never mind this one :)

PR #83, opened 2023-05-24
Make RuntimeMetricsTelemetry fields public

liustanley:
### What does this PR do?

Makes RuntimeMetricsTelemetry fields public so that we can access them from datadog-agent or opentelemetry-collector-contrib

### DIFF
--- a/pkg/otlp/metrics/metrics_translator.go
+++ b/pkg/otlp/metrics/metrics_translator.go
@@ -489,7 +489,7 @@ func (t *Translator) source(m pcommon.Map) (source.Source, error) {
 	return src, nil
 }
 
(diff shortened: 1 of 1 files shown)


pkg/otlp/metrics/metrics_translator.go
```diff
@@ -489,7 +489,7 @@ func (t *Translator) source(m pcommon.Map) (source.Source, error) {
 	return src, nil
 }
 
-// extractLanguageTag appends a new language tag to languageTags if a new language tag is found from the given name
+// extractLanguageTag appends a new language tag to LanguageTags if a new language tag is found from the given name
```
songy23, line 492: This doesn't need to be changed.

songy23 (approved):

mackjmr (approved):
LGTM!

PR #87, opened 2023-05-31
.github: add changelog workflow to ensure changelogs are on each PR

gbbr:


.github/workflows/changelog.yml
```diff
@@ -0,0 +1,90 @@
+# This action requires that any PR targeting the main branch should add a
```
mx-psi, line 1:
```suggestion
# Copyright The OpenTelemetry Authors
# SPDX-License-Identifier: Apache-2.0
# Source: github.com/open-telemetry/opentelemetry-collector-contrib/blob/6c71da2/.github/workflows/changelog.yml

# This action requires that any PR targeting the main branch should add a
```

mx-psi (approved):

PR #88, opened 2023-05-31
[pkg/inframetadata] Create new inframetadata module

mx-psi:
### What does this PR do?

Introduces a new module, `pkg/inframetadata` for handling infrastructure list metadata through a new `Reporter` struct.

Copies over the payload definitions for host metadata from the Datadog exporter.

### Motivation

Part of ongoing effort to improve metadata on remote hosts.


pkg/inframetadata/internal/hostmap/constants.go
```diff
@@ -0,0 +1,36 @@
...
+// Custom attributes not in the OpenTelemetry specification
+const (
+	attributeKernelName    = "os.kernel.name"
+	attributeKernelRelease = "os.kernel.release"
+	attributeKernelVersion = "os.kernel.version"
+)
+
+// gohai platform subsection fields.
+const (
+	platformHostname         = "hostname"
```
gbbr, line 16: To be consistent with the above, I'd prefix these with `field` too.

pkg/inframetadata/internal/hostmap/hostmap.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/DataDog/opentelemetry-mapping-go/pkg/inframetadata/gohai"
+	"github.com/DataDog/opentelemetry-mapping-go/pkg/inframetadata/payload"
+)
+
+// HostMap stores a map from hostnames to host metadata payloads.
+// Host metadata payloads can be updated with the information from a given resource.
+// At any time, the internal map can be extracted and the map will be cleared out.
+type HostMap struct {
+	// mutex for the host map.
+	mutex sync.Mutex
```
gbbr, line 19: This is generally called `mu`.

pkg/inframetadata/internal/hostmap/hostmap.go
```diff
@@ -0,0 +1,89 @@
...
+	"sync"
+
+	"go.opentelemetry.io/collector/pdata/pcommon"
+	"go.uber.org/multierr"
+
+	"github.com/DataDog/opentelemetry-mapping-go/pkg/inframetadata/gohai"
+	"github.com/DataDog/opentelemetry-mapping-go/pkg/inframetadata/payload"
+)
+
+// HostMap stores a map from hostnames to host metadata payloads.
```
gbbr, line 14:
```suggestion
// HostMap maps hostnames to host metadata payloads.
```
The rest seems like documentation for its methods. It is quite a simple API, so viewing this within the context of "godoc" will very easily reveal all that it's about. It only has 2 methods.

pkg/inframetadata/internal/hostmap/hostmap.go
```diff
@@ -0,0 +1,89 @@
...
+//
+// Partial modifications will still be applied even with non-fatal errors.
+func (m *HostMap) Update(host string, res pcommon.Resource) (changed bool, err error) {
+	hm := payload.HostMetadata{
+		Flavor:  "otelcol-contrib",
+		Meta:    &payload.Meta{},
+		Tags:    &payload.HostTags{},
+		Payload: gohai.NewEmpty(),
+	}
+	updater := newUpdater(res)
```
gbbr, line 45: I'd just use `u`.

pkg/inframetadata/internal/hostmap/hostmap.go
```diff
@@ -0,0 +1,89 @@
...
+	m.mutex.Lock()
+	defer m.mutex.Unlock()
+
+	var existingHost bool
+	if oldHM, ok := m.hosts[host]; ok {
```
gbbr, line 50:
```suggestion
	if old, ok := m.hosts[host]; ok {
```
Is `old` not enough?

pkg/inframetadata/internal/hostmap/hostmap.go
```diff
@@ -0,0 +1,89 @@
...
+		Flavor:  "otelcol-contrib",
+		Meta:    &payload.Meta{},
+		Tags:    &payload.HostTags{},
+		Payload: gohai.NewEmpty(),
+	}
+	updater := newUpdater(res)
+	m.mutex.Lock()
+	defer m.mutex.Unlock()
+
+	var existingHost bool
```
gbbr, line 49: Along the same lines as before ([naming things](https://datadoghq.atlassian.net/wiki/spaces/OTEL/pages/2580119947/Go+Coding+Guidelines#Naming-things)), surely `found` or `exists` should be enough.

pkg/inframetadata/gohai/gohai.go
```diff
@@ -0,0 +1,68 @@
...
+	if err != nil {
+		return []byte(""), err
+	}
+	return doubleMarshaledPayload, nil
+}
+
+// UnmarshalJSON implements the json.Unmarshaler interface.
+// Unmarshals the passed bytes twice (first to a string, then to gohai.Gohai)
+func (m *gohaiMarshaler) UnmarshalJSON(bytes []byte) error {
+	firstUnmarshall := ""
```
gbbr, line 49:
```suggestion
	var out string
```
Or feel free to name it otherwise, but I think we can use the zero-values.

pkg/inframetadata/gohai/gohai.go
```diff
@@ -0,0 +1,68 @@
...
+	err := json.Unmarshal(bytes, &firstUnmarshall)
+	if err != nil {
+		return err
+	}
+
+	err = json.Unmarshal([]byte(firstUnmarshall), &(m.Gohai))
+	return err
```
gbbr, line 56:
```suggestion
	return json.Unmarshal([]byte(firstUnmarshall), &(m.Gohai))
```

pkg/inframetadata/internal/hostmap/hostmap.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+// Update the information about a given host by providing a resource.
+// The function reports:
+//   - Whether the information about the `host` has changed
+//   - Any non-fatal errors that may have occurred during the update
+//
+// Partial modifications will still be applied even with non-fatal errors.
+func (m *HostMap) Update(host string, res pcommon.Resource) (changed bool, err error) {
+	hm := payload.HostMetadata{
```
gbbr, line 39: Many times metadata is called `md` (e.g. in gRPC package).

pkg/inframetadata/internal/hostmap/updater.go
```diff
@@ -0,0 +1,89 @@
...
+type updater struct {
+	// changed records if there have been any changes.
+	changed bool
+	// attrs stores attributes to use for update of the payload.
+	attrs pcommon.Map
+	// err records any non-fatal errors.
+	err error
+}
+
+// newUpdater creates a new Updater.
```
gbbr, line 21: What is `res` used for?

pkg/inframetadata/internal/hostmap/updater.go
```diff
@@ -0,0 +1,89 @@
...
+func newUpdater(res pcommon.Resource) *updater {
+	return &updater{
+		attrs: res.Attributes(),
+	}
+}
+
+// FromValue sets field from value and records changes.
+func (m *updater) FromValue(field *string, value string) {
```
gbbr, line 29: Why is `field` a pointer?

pkg/inframetadata/reporter.go
```diff
@@ -0,0 +1,149 @@
...
+		case <-r.closeCh:
+			cancel()
+			r.logger.Info("Stopped reporter")
+			return nil
+		}
+	}
+}
+
+// Stop the reporter.
+func (r *Reporter) Stop(_ context.Context) {
```
gbbr, line 147: Why do we need the argument?

pkg/inframetadata/reporter.go
```diff
@@ -0,0 +1,149 @@
...
+		hostMap: hostMap,
+		pusher:  pusher,
+		closeCh: make(chan struct{}),
+		ticker:  time.NewTicker(period),
+	}, nil
+}
+
+// checkResource to see if it should be used by default.
+// A resource is usable if 'AttributeDatadogHostUseAsMetadata' is true or shouldUseByDefault is true.
+func checkResource(res pcommon.Resource) (bool, error) {
```
gbbr, line 85: Why not call the function `shouldUse` or even better: `isHostmetadata`? And the variable simply `ok`?

pkg/inframetadata/reporter.go
```diff
@@ -0,0 +1,149 @@
...
+	shouldUse := shouldUseByDefault
```
gbbr, line 86: The fact that `shouldUseByDefault` is a constant somewhere else and you have to go there to figure it out is in my opinion not necessary.

pkg/inframetadata/reporter.go
```diff
@@ -0,0 +1,149 @@
...
+	if val, ok := res.Attributes().Get(AttributeDatadogHostUseAsMetadata); ok {
+		if val.Type() != pcommon.ValueTypeBool {
+			return false, fmt.Errorf("%q has type %q, expected \"Bool\"", AttributeDatadogHostUseAsMetadata, val.Type())
```
gbbr, line 89: I'm wondering if we can simplify the code here and simply have a fall back where we take the string representation of the attribute (when it's not bool) and if it's [parsable](https://pkg.go.dev/strconv#ParseBool) use that. If there is an error, we could simply log it here.

We could then get rid of the returned error value here, and also in `Consume`. Is it important for consume to return an error in this case? Isn't logging enough? That's what happens ultimately with the error anyway isn't it?

pkg/inframetadata/internal/hostmap/hostmap.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	m.hosts[host] = hm
+	changed = existingHost && updater.HasChanged()
+	err = multierr.Append(err, updater.Error())
+	return
+}
+
+// Extract all the host metadata payloads and clear them from the HostMap.
+func (m *HostMap) Extract() map[string]payload.HostMetadata {
```
gbbr, line 83: How about `AsMap`, `Copy`, or `Snapshot` (the latter indicating that it's just a current copy)? I liked those methods in the OTEL APIs (e.g. `AsRaw`)

pkg/inframetadata/gohai/gohai.go
```diff
@@ -0,0 +1,68 @@
+// Copyright The OpenTelemetry Authors
```
gbbr, line 1: I don't feel like this qualifies as a separate package. Actually, I think all you need is one package for all of this. None of these packages need to be re-used elsewhere, nor are they individually useful. A package is not meant for separating logic, it's supposed to have a self-contained scope and be usable by at least two other packages.

pkg/inframetadata/internal/hostmap/hostmap.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	updater.FromValue(&hm.InternalHostname, host)
+
+	// Meta section
+	updater.FromFunc(&hm.Meta.InstanceID,
+		func(m pcommon.Map) (string, bool, error) {
+			return "", false, fmt.Errorf("not implemented: InstanceID")
+		},
+	)
```
gbbr, line 62:
```suggestion
	updater.FromFunc(&hm.Meta.InstanceID, notImplemented("InstanceID"))
```
This would make things a bit cleaner.

pkg/inframetadata/internal/hostmap/hostmap.go
```diff
@@ -0,0 +1,89 @@
...
+			return "", false, fmt.Errorf("not implemented: EC2Hostname")
+		},
+	)
+	updater.FromValue(&hm.Meta.Hostname, host)
+
+	// Gohai - Platform
+	updater.FromValue(hm.Gohai.Gohai.Platform.(map[string]*string)["hostname"], host)
+	for field, attribute := range platformAttributesMap {
+		updater.FromAttributeToMap(hm.Gohai.Gohai.Platform.(map[string]string), field, attribute)
+	}
```
gbbr, line 74: This part needs some inline documentation and clarification. It is too hard to understand what's going on here. I suspect we can make some improvements on this end.

I feel like the `updater` could be simplified by simply taking an `hm` and a `host` and do everything under the hood, but I don't understand what's going on here that well yet to be able to suggest improvements. Maybe later on in a follow up review?

pkg/inframetadata/payload/payload.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/DataDog/opentelemetry-mapping-go/pkg/inframetadata/gohai"
+)
+
+// HostMetadata includes metadata about the host tags,
+// host aliases and identifies the host as an OpenTelemetry host
+type HostMetadata struct {
+	// Meta includes metadata about the host.
+	Meta *Meta `json:"meta"`
+
+	// InternalHostname is the canonical hostname
```
gbbr, line 13: What is the difference between this hostname and the meta hostname?

pkg/inframetadata/internal/hostmap/hostmap.go
```diff
@@ -0,0 +1,89 @@
...
+	)
+	updater.FromFunc(&hm.Meta.EC2Hostname,
+		func(m pcommon.Map) (string, bool, error) {
+			return "", false, fmt.Errorf("not implemented: EC2Hostname")
+		},
+	)
+	updater.FromValue(&hm.Meta.Hostname, host)
+
+	// Gohai - Platform
+	updater.FromValue(hm.Gohai.Gohai.Platform.(map[string]*string)["hostname"], host)
```
gbbr, line 71: This cast needs to be improved somehow or isolated away as a method on the `Gohai` type. Also, I see possibilities of panics here so we need to make sure that both the cast worked and that the map is not nil.

gbbr (commented):
I hope this helps.

pkg/inframetadata/gohai/gohai.go
```diff
@@ -0,0 +1,68 @@
...
+	Memory     interface{} `json:"memory"`
+	Network    interface{} `json:"network"`
+	Platform   interface{} `json:"platform"`
+}
+
+// Payload handles the JSON unmarshalling of the metadata payload
+// As weird as it sounds, in the v5 payload the value of the "gohai" field
+// is a JSON-formatted string. So this struct contains a MarshaledGohaiPayload
+// which will be marshaled as a JSON-formatted string.
+type Payload struct {
```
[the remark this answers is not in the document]
mx-psi: Yeah, I don't get that either myself :) But as I said above, this is copy-pasta, so let's do it separately

pkg/inframetadata/README.md
```diff
@@ -0,0 +1,18 @@
+# `pkg/inframetadata`
+
+`pkg/inframetadata` handles host metadata and infrastructure list related features. It stores the host metadata and gohai payload definitions as well as the `Reporter` implementation.
+
+## `inframetadata.Reporter` implementation
```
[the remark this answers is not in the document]
mx-psi: This is already on the godoc, I wanted to do a summary. I will move it to the `doc.go` file

pkg/inframetadata/gohai/processes.go
```diff
@@ -0,0 +1,10 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package gohai
+
+// ProcessesPayload handles the JSON unmarshalling
```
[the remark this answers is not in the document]
mx-psi: This is copy-paste as well :)

mx-psi:
Note for reviewers: the `pkg/inframetadata/gohai` and `pkg/inframetadata/payload` packages have been copied over from the Datadog exporter codebase. I would like to keep backwards-incompatible changes on this PR or improvements on these packages that are not strictly related to the functionality on this PR out of scope, we can handle them in separate, future PRs once this one has landed.

pkg/inframetadata/internal/hostmap/updater.go
mx-psi: Because we want to update it, not just use its value.

pkg/inframetadata/reporter.go
mx-psi: We don't need it now, I am just trying to be consistent with the usual component interface within the OpenTelemetry Collector. But we can leave it out for now if you think it is unnecessary and change the interface in the future.

pkg/inframetadata/reporter.go
mx-psi: > I'm wondering if we can simplify the code here and simply have a fall back where we take the string representation of the attribute (when it's not bool) and if it's [parsable](https://pkg.go.dev/strconv#ParseBool) use that. If there is an error, we could simply log it here.

I chose a more conservative approach since it's not clear to me what the behavior should be with a string. Should we accept `1` as true? What about `True`? `TRUE`? `yes`? Is it going to be confusing to the user if we don't accept any of those?

> We could then get rid of the returned error value here, and also in Consume. Is it important for consume to return an error in this case? Isn't logging enough? That's what happens ultimately with the error anyway isn't it?

Logging is what ultimately will happen in this case, yes, but it's not necessarily the case if we reuse this code elsewhere (e.g. in the backend). I think returning an error makes it more general.

pkg/inframetadata/internal/hostmap/hostmap.go
mx-psi: I don't think any of those names make it clear that we are clearing the map when calling this method, which is a very important aspect of it

pkg/inframetadata/internal/hostmap/hostmap.go
mx-psi: I think I will implement these on this PR, so I will just keep it like this until I do so.

pkg/inframetadata/README.md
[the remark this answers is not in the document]
gbbr: It looks like the new documentation generator picks up the readme too, so no need to.

pkg/inframetadata/gohai/gohai.go
```diff
@@ -0,0 +1,68 @@
...
+type Gohai struct {
+	CPU        interface{} `json:"cpu"`
+	FileSystem interface{} `json:"filesystem"`
+	Memory     interface{} `json:"memory"`
+	Network    interface{} `json:"network"`
+	Platform   interface{} `json:"platform"`
+}
+
+// Payload handles the JSON unmarshalling of the metadata payload
+// As weird as it sounds, in the v5 payload the value of the "gohai" field
```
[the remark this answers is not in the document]
gbbr: I don't understand why that is an excuse to leave the code in a bad shape. If it's reused, why isn't it imported from there? And if it's copied, why can't it be rewritten and improved? It's just a few structures. Why are we copying terrible documentation and using the fact that it's a copy as an excuse to not improve it?

Furthermore, these are just structs that are marshalled (as far as I can see) for an API somewhere. They might as well be rewritten from zero here. What is the point of perpetuating this bad code everywhere?

gbbr:
> Note for reviewers: the pkg/inframetadata/gohai and pkg/inframetadata/payload packages have been copied over from the Datadog exporter codebase. I would like to keep backwards-incompatible changes on this PR or improvements on these packages that are not strictly related to the functionality on this PR out of scope, we can handle them in separate, future PRs once this one has landed.

Can you please explain how improving the copy of a struct which is merely going to be marshalled a backwards incompatible change? I stand by my point in https://github.com/DataDog/opentelemetry-mapping-go/pull/88#discussion_r1219635078.

In my opinion, we should either:
a) find a way to reuse the bad code
b) copy it but then rewrite it correctly

pkg/inframetadata/gohai/gohai.go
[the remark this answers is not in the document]
gbbr: I don't understand why that is an excuse to leave the code in a bad shape. If it's reused, why isn't it imported from there? And if it's copied, why can't it be rewritten and improved? It's just a few structures. Why are we copying terrible documentation and using the fact that it's a copy as an excuse to not improve it?

Furthermore, these are just structs that are marshalled (as far as I can see) for an API somewhere. They might as well be rewritten from zero here. What is the point of perpetuating this bad code everywhere?

How is someone coming to read this code ever going to figure out what all of this is? What is a gohai? IMHO this is not acceptable code quality. Having a tech debt somewhere and reusing that is fine, but perpetuating it and multiplying it everywhere and not improving it with that excuse is just wrong. Then, we're using more excuses due to it to allow hacks like [this](https://github.com/DataDog/opentelemetry-mapping-go/pull/88/files#r1211750225).

gbbr (changes requested):
Would be helpful if you could explain why the files are copied, and why the copy can not be changed. IMHO this is unacceptable quality. If I, as a Senior eng. can't figure out this code, how will other contributors?

mx-psi:
From offline chat with @gbbr: I will add a header on all copied files about gohai, explaining:
- where it comes from
- what work is going on the refactor
- why it's copied
- a hint about any modifications that are made

We also want to add some modifications if we expect them to be useful after the Agent shared components refactor is completed

pkg/inframetadata/README.md
```diff
@@ -0,0 +1,18 @@
+# `pkg/inframetadata`
+
+`pkg/inframetadata` handles host metadata and infrastructure list related features. It stores the host metadata and gohai payload definitions as well as the `Reporter` implementation.
```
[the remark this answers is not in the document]
mx-psi: Moved to doc.go

pkg/inframetadata/gohai/gohai.go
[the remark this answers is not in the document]
mx-psi: This was addressed on #105

pkg/inframetadata/internal/hostmap/constants.go
mx-psi: Done

pkg/inframetadata/internal/hostmap/hostmap.go
mx-psi: Done

pkg/inframetadata/internal/hostmap/hostmap.go
mx-psi: Removed the extra documentation; applied the suggested change

pkg/inframetadata/internal/hostmap/hostmap.go
mx-psi: Renamed

pkg/inframetadata/internal/hostmap/hostmap.go
mx-psi: Renamed

pkg/inframetadata/internal/hostmap/hostmap.go
mx-psi: Renamed

pkg/inframetadata/internal/hostmap/hostmap.go
mx-psi: Renamed

pkg/inframetadata/internal/hostmap/hostmap.go
mx-psi: Renamed to `Flush`

pkg/inframetadata/internal/hostmap/hostmap.go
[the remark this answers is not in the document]
mx-psi: I made this a field of the `HostMap`. Since it is only modified under a mutex, we can clear and reuse the same `updater` for each call. If you think there's a better approach I am all ears :)

pkg/inframetadata/internal/hostmap/updater.go
mx-psi: We read its contents for updates; the `updater` handles extracting values from a resource and checking their types

pkg/inframetadata/payload/payload.go
mx-psi: Addressed on #105

pkg/inframetadata/gohai/gohai.go
mx-psi: Done

pkg/inframetadata/reporter.go
mx-psi: I feel like we don't have to take this decision now, expanding the set of values we support is a backwards compatible change and we can do it once we have had some users try this out. Put it simply: if we change it now, we commit to supporting the whole set of values forever, if we don't we can always change it later.

Do you think this is important to support now (e.g. do you feel like this would be confusing as is for the user)?

pkg/inframetadata/internal/hostmap/hostmap.go
```diff
@@ -0,0 +1,89 @@
...
+		hosts: make(map[string]payload.HostMetadata),
+	}, nil
+}
+
+// Update the information about a given host by providing a resource.
+// The function reports:
+//   - Whether the information about the `host` has changed
+//   - Any non-fatal errors that may have occurred during the update
+//
+// Partial modifications will still be applied even with non-fatal errors.
```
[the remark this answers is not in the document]
mx-psi: Imagine you get a payload with the following fields:

- `host.name` with a string value `myhostname`
- `os.kernel` with a string value `Linux`
- `host.arch` with a string value `x86_64`
- `os.description` with a boolean value of `true` (this is wrong, should be a string)

What should we do? I see two options:

1. Update the fields that have the correct values (i.e. all of them but `os.description`), and return an error signaling that the type of `os.description` is wrong
2. Don't update any fields, since one of them is wrong, and return an error signaling that the type of `os.description` is wrong

Option (1) is what we are currently doing (the `os.description` type mismatch is non-fatal), option (2) would be what we do if we restore the original value.

I feel like option (1) is better in that the error is local to `os.description` and so the user experience is only partially degraded (one field missing vs. all fields missing). Do you think we should do option (2)? Or maybe the comment was unclear?

pkg/inframetadata/internal/hostmap/hostmap.go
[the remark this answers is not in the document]
mx-psi: I decided to get rid of the updater entirely

pkg/inframetadata/internal/hostmap/hostmap.go
mx-psi: I got rid of the updater, it's more verbose but I think it was worth it

pkg/inframetadata/internal/hostmap/updater.go
mx-psi: not using updater anymore

mx-psi:
Still needs tests (will get to that tomorrow), but I addressed most comments and left a reply otherwise

pkg/inframetadata/internal/hostmap/hostmap.go
```diff
@@ -0,0 +1,143 @@
...
+	if EC2Hostname, ok, fieldErr := getEC2Hostname(res.Attributes()); fieldErr != nil {
+		err = multierr.Append(err, fieldErr)
+	} else if ok {
+		old := md.Meta.EC2Hostname
+		changed = changed || old != EC2Hostname
+		md.Meta.EC2Hostname = EC2Hostname
+	}
+
+	// Gohai - Platform
+	md.Gohai.Gohai.Platform.(map[string]string)["hostname"] = host
```
gbbr, line 119: I thought you agreed to add a method for accessing this map to avoid this ugliness :)

pkg/inframetadata/internal/hostmap/hostmap.go
```diff
@@ -0,0 +1,143 @@
...
+	if err != nil {
+		return false, err
+	} else if !ok {
+		// no cloud provider field
+		return false, nil
+	}
+	return cloudProvider == conventions.AttributeCloudProviderAWS, nil
+}
+
+func getInstanceID(m pcommon.Map) (string, bool, error) {
```
gbbr, line 61: Can we void the `get*` prefix? https://go.dev/doc/effective_go#Getters

Also, can you please document these and their return values?

pkg/inframetadata/internal/hostmap/hostmap.go
```diff
@@ -0,0 +1,143 @@
...
+	if !ok {
+		// Field not available, don't update but don't fail either
+		return "", false, nil
+	}
+
+	if val.Type() != pcommon.ValueTypeStr {
+		return "", false, fmt.Errorf("%q has type %q, expected type \"Str\" instead", key, val.Type())
+	}
+
+	return val.Str(), true, nil
```
gbbr, line 47: I prefer the style you've used elsewhere too (and as is in our style guidelines), with no empty lines around curly braces.

pkg/inframetadata/internal/hostmap/hostmap.go
```diff
@@ -0,0 +1,143 @@
...
+// HostMap maps from hostnames to host metadata payloads.
+type HostMap struct {
+	// mu is the mutex for the host map and updater.
+	mu sync.Mutex
+	// hosts map
+	hosts map[string]payload.HostMetadata
+}
+
+// New creates a new HostMap.
+func New() (*HostMap, error) {
```
gbbr, line 29: I think you can drop the error return right?

pkg/inframetadata/internal/hostmap/hostmap.go
```diff
@@ -0,0 +1,143 @@
...
+	return &HostMap{
+		mu:    sync.Mutex{},
```
gbbr, line 31: You can remove this line. The mutex will be initialised as its zero value.

pkg/inframetadata/reporter.go
gbbr: If it's not needed for anything, I don't think we should have it.

pkg/inframetadata/internal/hostmap/updater.go
mx-psi: not using the updater anymore, marking this as solved

pkg/inframetadata/internal/hostmap/hostmap.go
mx-psi: Dropped and documented

pkg/inframetadata/internal/hostmap/hostmap.go
mx-psi: Done

pkg/inframetadata/internal/hostmap/constants.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+// Custom attributes not in the OpenTelemetry specification
+const (
+	attributeKernelName    = "os.kernel.name"
+	attributeKernelRelease = "os.kernel.release"
+	attributeKernelVersion = "os.kernel.version"
+)
+
+// gohai fieldPlatform subsection fields.
```
gbbr, line 19: Please extend the "gohai fieldPlatform" thing. Maybe you mean to say:
```go
// This set of constants represent fields in the Gohai payload's Platform field.
```

pkg/inframetadata/internal/hostmap/hostmap.go
gbbr: I was thinking a getter/setter, e.g.:
```go
md.SetPlatform(k, v)
```
But the way you did it is fine too. Except you aren't using it here.

pkg/inframetadata/internal/hostmap/hostmap.go
```diff
@@ -0,0 +1,157 @@
...
+	if old, ok := m.hosts[host]; ok {
+		found = true
+		md = old
+	}
+
+	md.InternalHostname = host
+	md.Meta.Hostname = host
+
+	// InstanceID field
+	if instanceId, ok, fieldErr := instanceID(res.Attributes()); fieldErr != nil {
```
gbbr, line 115: Is there any chance we could follow the naming conventions laid out in our style guidelines? https://datadoghq.atlassian.net/wiki/spaces/OTEL/pages/2580119947/Go+Coding+Guidelines#Naming-things

I feel like this "long variable names" style is more specific to languages like Java and less so to Go. I am aware this habit is prevalent in the OpenTelemetry repository, but I don't agree we should write less idiomatic Go code because that org. is doing so.

However, happy to debate and change the style guide at any time.

mx-psi: Do you have a specific name suggestion for this one in particular? I can't think of a good short name.

pkg/inframetadata/internal/hostmap/constants.go
mx-psi: Done

pkg/inframetadata/internal/hostmap/hostmap.go
mx-psi: oops, using it now

pkg/inframetadata/internal/hostmap/hostmap.go
mx-psi: I think this is addressed by the `Platform` method

pkg/inframetadata/internal/hostmap/hostmap.go
gbbr:
```
if iid, ok, err2 ...
```
So I guess `id`, `iid`, `v`, all work here...

Is something I saw outside in the while. `err2` being "error nested at level 2". My comment was more general for everywhere.

mx-psi: Done

mx-psi: Done for instance id and EC2 hostname. I think for the most part the rest have been changed on previous commits, but feel free to reopen if there are any missing

pkg/inframetadata/internal/hostmap/hostmap.go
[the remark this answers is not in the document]
songy23: Thanks, I think the current behavior makes more sense. Maybe update the comment to also mention order doesn't matter and all fields without errors will be applied. I.e. in your example, if it's `os.description`, `host.name`, `host.arc`, `os.kernel` it will yield the same result.

pkg/inframetadata/gohai/gohai.go
```diff
@@ -59,13 +64,13 @@ func (m gohaiMarshaler) MarshalJSON() ([]byte, error) {
...
-	firstUnmarshall := ""
-	err := json.Unmarshal(bytes, &firstUnmarshall)
+	var out string
+	err := json.Unmarshal(bytes, &out)
 	if err != nil {
 		return err
 	}
 
-	err = json.Unmarshal([]byte(firstUnmarshall), &(m.Gohai))
+	err = json.Unmarshal([]byte(out), &(m.Gohai))
```
songy23, line 73: For my own understanding: why cannot it unmarshal `bytes` directly?

pkg/inframetadata/reporter.go
```diff
@@ -0,0 +1,158 @@
...
+			zap.String("host", src.Identifier), zap.Any("attributes", res.Attributes()),
+		)
+	}
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// Run the reporter to periodically exporte
```
songy23, line 129:
```suggestion
// Run the reporter to periodically export
```

pkg/inframetadata/gohai/gohai.go
mx-psi: This comment adds some context https://github.com/DataDog/opentelemetry-mapping-go/blob/1e4c8cc3796bc118ddc2ae31bee54ac60f6bd910/pkg/inframetadata/gohai/gohai.go#L32-L34 for some reason the `gohai` field does not contain a JSON struct, but rather a serialization of a JSON struct into a string.

pkg/inframetadata/internal/hostmap/hostmap.go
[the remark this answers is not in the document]
mx-psi: Updated comment to clarify: https://github.com/DataDog/opentelemetry-mapping-go/blob/aec0871cf47d1dac76eebce1378b034ddce47c99/pkg/inframetadata/internal/hostmap/hostmap.go#L94-L100

pkg/inframetadata/gohai/gohai.go
songy23: Maybe add some comments here as people who read these lines may have the same doubt

pkg/inframetadata/internal/hostmap/hostmap_test.go
```diff
@@ -0,0 +1,139 @@
...
+				conventions.AttributeHostID:        "host-1-hostid",
+				conventions.AttributeHostName:      "host-1-hostname",
+				conventions.AttributeOSDescription: true, // wrong type
+				conventions.AttributeHostArch:      conventions.AttributeHostArchAMD64,
+				attributeKernelName:                "GNU/Linux",
+				attributeKernelRelease:             "5.19.0-43-generic",
+				attributeKernelVersion:             "#82~18.04.1-Ubuntu SMP Fri Apr 16 15:10:02 UTC 2021", // changed
+			},
+			expectedChanged: true,
+			expectedErr:     "\"os.description\" has type \"Bool\", expected type \"Str\" instead",
```
songy23, line 64: nit: consider adding a test with multiple errors that are all reported

songy23 (approved):
Minor comments, otherwise LGTM

pkg/inframetadata/gohai/gohai.go
gbbr, line 73: Why not just return this line?
```suggestion
	return json.Unmarshal([]byte(out), &(m.Gohai))
```

gbbr (approved):

pkg/inframetadata/gohai/gohai.go
mx-psi: Added a couple comments

mx-psi: Done

pkg/inframetadata/internal/hostmap/hostmap_test.go
mx-psi: Done

PR #89, opened 2023-05-31
Add pkg/otlp/logs for OTLP to DD log translation

IbraheemA:
Copy over `opentelemetry-collector-contrib/exporter/datadogexporter/internal/logs/translator.go`, which translates logs that were ingested from OTel from `plog.LogRecord` to `datadogV2.HTTPLogItem`.

This will be imported in datadog-agent for use in the OTLP log ingestion pipeline, and existing uses of the translator in `opentelemetry-collector-contrib` will be replaced with calls to this module.


songy23 (commented):
Please add a changelog entry

.chloggen/otel-529-copy-otlp-translator-code.yaml
```diff
@@ -0,0 +1,16 @@
...
+change_type: new_component
+
+# The name of the component (e.g. pkg/quantile)
+component: pkg/otlp/logs
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Create new `pkg/otlp/logs` module for OTLP to DD log translation
+
+# The PR related to this change
+issues: []
```
songy23, line 11:
```
Error: specify one or more issues #'s
```

Please file an issue in this repo and add here.

songy23 (approved):

songy23:
cc @mx-psi we may want to skip `test / breaking-changes` for new components

mackjmr (approved):
It may also be good to add the `doc.go` ([link](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/datadogexporter/internal/logs/doc.go)) which contains a description of what the logs package does. Only the first part of the documentation is relevant, as the sender part stayed in opentelemetry-collector-contrib.

Otherwise LGTM!

pkg/otlp/logs/go.mod
```diff
@@ -0,0 +1,35 @@
+module github.com/DataDog/opentelemetry-mapping-go/pkg/otlp/logs
+
+go 1.19
+
+require (
+	github.com/DataDog/datadog-api-client-go/v2 v2.13.0
+	github.com/DataDog/opentelemetry-mapping-go/pkg/otlp/attributes v0.2.3
```
mx-psi, line 7: You need to add a replace statement for this, similar to https://github.com/DataDog/opentelemetry-mapping-go/blob/efedb643325cbc0ab38912cb29bab350c320cbbf/pkg/otlp/metrics/go.mod#L44

This will ensure that we are testing with the latest version of the code instead of the latest released version

PR #101, opened 2023-06-14
pkg/otlp/metrics: add remapping code

### LINKED ISSUE #102: pkg/otlp/metrics: move remapping code from exporter
This change involves moving the metric [remapping code](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.79.0/exporter/datadogexporter/internal/metrics/system.go) from the Datadog Exporter to this repository. The operation must be off by default so that it does not affect nor break the Datadog Agent which does not require this path. It should only be enabled by means of a new translator config option.

Internal ref: [OTEL-554](https://datadoghq.atlassian.net/browse/OTEL-554)

[OTEL-554]: https://datadoghq.atlassian.net/browse/OTEL-554?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

gbbr:
This change adds the processing code for extracting Datadog metrics from container.* and system.* metrics with the intention of centralising this logic to be re-used in the backend intake.

The operation is restricted by the newly added WithRemapping translator configuration option, so that it does not affect the translator's usage as part of the Datadog Agent, where no remapping is necessary because the agent creates the necessary metrics.

Closes #102

### DIFF
--- a/pkg/otlp/metrics/config.go
+++ b/pkg/otlp/metrics/config.go
@@ -51,6 +51,9 @@ type TranslatorOption func(*translatorConfig) error
 // WithRemapping specifies that certain OTEL metrics (such as container.* and system.*) need to be
 // remapped to their Datadog counterparts because they will not be available otherwise. This happens
 // in situations when the translator is running as part of a Collector without the Datadog Agent.
+//
+// Do note that in some scenarios this process renames certain metrics (such as for example prefixing
+// system.* and process.* metrics with the otel.* namespace).
 func WithRemapping() TranslatorOption {
 	return func(t *translatorConfig) error {
 		t.withRemapping = true
--- a/pkg/otlp/metrics/config.go
+++ b/pkg/otlp/metrics/config.go
@@ -33,6 +33,11 @@ type translatorConfig struct {
 	InstrumentationLibraryMetadataAsTags bool
 	InstrumentationScopeMetadataAsTags   bool
 
+	// withRemapping reports whether certain metrics that are only available when using
+	// the Datadog Agent should be obtained by remapping from OTEL counterparts (e.g.
+	// container.* and system.* metrics).
+	withRemapping bool
+
 	// cache configuration
 	sweepInterval int64
 	deltaTTL      int64
@@ -42,6 +47,16 @@ type translatorConfig struct {
 
 // TranslatorOption is a translator creation option.
 type TranslatorOption func(*translatorConfig) error
+
+// WithRemapping specifies that certain OTEL metrics (such as container.* and system.*) need to be
+// remapped to their Datadog counterparts because they will not be available otherwise. This happens
+// in situations when the translator is running as part of a Collector without the Datadog Agent.
+func WithRemapping() TranslatorOption {
+	return func(t *translatorConfig) error {
+		t.withRemapping = true
+		return nil
+	}
+}
 
 // WithDeltaTTL sets the delta TTL for cumulative metrics datapoints.
 // By default, 3600 seconds are used.
--- a/pkg/otlp/metrics/metrics_remapping.go
+++ b/pkg/otlp/metrics/metrics_remapping.go
@@ -48,15 +48,11 @@ func remapSystemMetrics(all pmetric.MetricSlice, m pmetric.Metric) {
 	case "system.cpu.load_average.15m":
 		copyMetric(all, m, "system.load.15", 1)
 	case "system.cpu.utilization":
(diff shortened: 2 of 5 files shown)


pkg/otlp/metrics/metrics_remapping.go
```diff
@@ -0,0 +1,193 @@
...
+	if len(tags) == 0 {
+		return true
+	}
+	attr := point.Attributes()
+	for _, tag := range tags {
+		v, ok := attr.Get(tag.K)
+		if !ok {
+			continue
+		}
+		if v.Str() == tag.V {
```
mx-psi, line 188: This panics if `v.Type() != pcommon.ValueTypeStr`. I think we should check for this first (needs a new import):
```suggestion
		if v.Type() == pcommon.ValueTypeStr  && v.Str() == tag.V {
```

pkg/otlp/metrics/config.go
```diff
@@ -43,6 +48,16 @@ type translatorConfig struct {
 // TranslatorOption is a translator creation option.
 type TranslatorOption func(*translatorConfig) error
 
+// WithRemapping specifies that certain OTEL metrics (such as container.* and system.*) need to be
+// remapped to their Datadog counterparts because they will not be available otherwise. This happens
+// in situations when the translator is running as part of a Collector without the Datadog Agent.
```
mx-psi, line 53: Using this option means that the pdata structs passed to the Translator cannot be used concurrently by other code since we modify them. Should we add a comment about this?

pkg/otlp/metrics/metrics_remapping.go
mx-psi: Ah, right, my bad, I got confused with how `pmetric.Metric` works. Marking this as solved

gbbr:
I guess this is ready for review and fully tested now, but there seems to have been a breaking change in a dependency which is causing the CI to fail oddly.

mx-psi:
> I guess this is ready for review and fully tested now, but there seems to have been a breaking change in a dependency which is causing the CI to fail oddly.

I think it will be fixed once you rebase/merge main

pkg/otlp/metrics/config.go
mx-psi: > Now that I think about it, it does prepend the otel.* prefix to some metrics, so it changes existing metrics (not just add new ones); maybe that is worth mentioning indeed.

Yeah, I was more worried about this, but it's true that the runtime metrics already do some modifications. I think we should mention the name change.

pkg/otlp/metrics/metrics_remapping.go
```diff
@@ -0,0 +1,194 @@
...
+
+// kv represents a key/value pair.
+type kv struct{ K, V string }
+
+// copyMetric copies metric m to dest. The new metric's name will be newname, and all of its datapoints will
+// be divided by div. If filter is provided, only the data points that have *either* of the specified string
+// attributes will be copied over. If the filtering results in no datapoints, no new metric is added to dest.
+//
+// Please note that copyMetric is restricted to the metric types Sum and Gauge.
+func copyMetric(dest pmetric.MetricSlice, m pmetric.Metric, newname string, div float64, filter ...kv) (pmetric.Metric, bool) {
```
songy23, line 137: May also want to document the returned `bool`

pkg/otlp/metrics/metrics_remapping_test.go
```diff
@@ -0,0 +1,447 @@
...
+	// point is a datapoint
+	type point struct {
+		// i defines a IntValue datapoint when non-zero
+		i int64
+		// f defines a DoubleValue datapoint when non-zero
+		f float64
+		// attrs specifies the raw attributes of the datapoint
+		attrs map[string]any
+	}
+	// m is a convenience function to create a new metric with the given name
```
songy23, line 36:
```suggestion
	// metric is a convenience function to create a new metric with the given name
```

pkg/otlp/metrics/metrics_remapping.go
gbbr: Indeed. I added a sentence about the return values.

pkg/otlp/metrics/metrics_remapping_test.go
gbbr: Thanks. I missed this.

mx-psi (approved):

PR #103, opened 2023-06-14
[otlp] update protobuf definitions to new location/version in datadog-agent repo

truthbk:
### What does this PR do?

This PR updates the protobuf imports such that the new definitions in `pkg/proto` in the agent repository are used. This includes some changes due to the gogoproto `nullable` type being removed from the definitions. 

Additionally, due to a chicken-egg problem, these changes are currently pointing to a WIP PR on the datadog-agent repository (PR to be posted soon). Unfortunately the current protobuf structure (definitions part of https://github.com/DataDog/datadog-agent) doesn't make it easy to make these updates. 

### Motivation

Refactor of protobuf definitions in https://github.com/DataDog/datadog-agent so that all definitions now live in `pkg/proto` as opposed to scattered all over the repository.


LICENSE-3rdparty.csv
```diff
@@ -145,7 +145,7 @@ pkg/otlp/attributes,google.golang.org/protobuf/types/descriptorpb,BSD-3-Clause,C
 pkg/otlp/attributes,google.golang.org/protobuf/types/known/anypb,BSD-3-Clause,Copyright (c) 2018 The Go Authors. All rights reserved
 pkg/otlp/attributes,google.golang.org/protobuf/types/known/durationpb,BSD-3-Clause,Copyright (c) 2018 The Go Authors. All rights reserved
 pkg/otlp/attributes,google.golang.org/protobuf/types/known/timestamppb,BSD-3-Clause,Copyright (c) 2018 The Go Authors. All rights reserved
-pkg/otlp/metrics,github.com/DataDog/datadog-agent/pkg/trace/pb,Apache-2.0,"Copyright 2016-present Datadog, Inc"
+pkg/otlp/metrics,github.com/DataDog/datadog-agent/pkg/proto/pbgo/trace,Apache-2.0,"Copyright 2016-present Datadog, Inc"
```
songy23, line 148: This is the only change needed in this file, please revert other changes.

.github/workflows/test.yaml
```diff
@@ -92,10 +92,10 @@ jobs:
         run: make tidy && git diff -s --exit-code || (echo "make tidy needed"; exit 1)
       - name: Check gofmt
         run: make fmt  && git diff -s --exit-code || (echo  "make fmt needed"; exit 1)
-      - name: Check licenses file
```
songy23, line 95: Please revert

truthbk: Yeah, was just checking the `fmt` stuff was 👍

songy23 (approved):

gbbr:
Can we revert this change? This package doesn't even exist on `main` (https://github.com/DataDog/datadog-agent/tree/main/pkg/proto/pbgo), and we can't update the exporter anymore to use this code because it depends on `pkg/trace` from the datadog-agent, which isn't updated.

songy23:
I believe this needs to be merged to break a circular dependency issue (?) in Agent because @truthbk needs a pseudo version of mapping go. But if this is blocking the upgrade in datadog exporter, I agree we should revert it for now.

gbbr:
Sure, but this can be referenced elsewhere as a commit from this (unmerged) PR so it should be fine for Jaime (I think?)

PR #105, opened 2023-06-16
[pkg/inframetadata] Create module and define payloads

mx-psi:
### What does this PR do?

Defines the host metadata and gohai payloads. This code is copied over from the Datadog exporter.

### Motivation

Split off from #88


gbbr (approved):
Thanks for the thorough documentation. This is great 👍🏻

songy23 (approved):

PR #107, opened 2023-06-16
Retract v0.4.0

gbbr:
This change retracts v0.4.0 and reverts #103.

#103 shouldn't have been released. It is blocking us from updating the contrib collector because pkg/trace in datadog-agent isn't updated with these changes. Furthermore, those changes aren't yet out.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,14 +6,7 @@
 
 ## v0.4.0
 
-### 🛑 Breaking changes 🛑
-
-- `pkg/otlp`: use new location for updated protobuf definitions in the datadog-agent repo, trace protobuf definitions specifically. (#103)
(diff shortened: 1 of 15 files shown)


CHANGELOG.md
```diff
@@ -6,14 +6,7 @@
 
 ## v0.4.0
 
-### 🛑 Breaking changes 🛑
```
mx-psi, line 9: Usually we should keep the changelog, the version is still available and if someone was already using it they can keep doing so if they want. In this instance I don't think anyone is using it so I will leave whether to bring the changelog back up to you :)

mx-psi (approved):

mx-psi:
cc @truthbk we are (temporarily) reverting your PR, it shouldn't have been merged until the `pkg/trace` changes were released. We can add it back once they have landed

songy23 (approved):
cc @truthbk

PR #114, opened 2023-06-28
Add support for OTel Python runtime metrics

liustanley:
### What does this PR do?

Adds mappings and prefixes for OTel Python runtime metrics. The v2 DD Python names are still undecided, so the DD prefixes in the mappings will change before merging.

### DIFF
--- a/.chloggen/stanley.liu_python-runtime-metrics.yaml
+++ b/.chloggen/stanley.liu_python-runtime-metrics.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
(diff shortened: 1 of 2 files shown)


liustanley:
> > The v2 DD Python names are still undecided, so the DD prefixes in the mappings will change before merging.
> 
> Should we mark this as draft then until this is decided? Or is this ready for review now?

This can be reviewed now, the only thing that will change are the metric names and I won't merge before updating them.

mx-psi (commented):
LGTM save for the metric names. Will actually approve once the names are ready!

majorgreys:
@liustanley Do we still want this?

liustanley:
> @liustanley Do we still want this?

These mappings are probably out of date, I'll close this PR but feel free to open a new one if you want to add Python runtime metric mappings.

PR #133, opened 2023-07-27
[chore] Shrink `Agent` struct from 104 bytes to 56

blt:
### What does this PR do?

This commit reorders the fields in the Agent struct to require less padding in that type, shrinking it by ~1/2. No change in behavior is intended.

### Motivation

This was discovered while investigating high memory use in the Agent under some circumstances, see SMP-664 and AML-825.


blt:
Apologies for the force-pushes. I intended to re-trigger CI. I don't have permissions to push branches to this repo directly, whence the fork to my personal account.

songy23 (approved):

PR #142, opened 2023-08-02
pkg/otlp/metrics: fix otel.* namespaced metrics

gbbr:
This change fixes a problem where sytem.* and process.* metrics were correctly namespaced with otel.*, but the original metric was not kept.

### DIFF
--- a/pkg/otlp/metrics/metrics_remapping.go
+++ b/pkg/otlp/metrics/metrics_remapping.go
@@ -70,7 +70,9 @@ func remapSystemMetrics(all pmetric.MetricSlice, m pmetric.Metric) {
 		copyMetric(all, m, "system.disk.in_use", 1)
 	}
 	// process.* and system.* metrics need to be prepended with the otel.* namespace
-	m.SetName("otel." + m.Name())
+	newm := all.AppendEmpty()
+	m.CopyTo(newm)
+	newm.SetName("otel." + m.Name())
 }
 
 // remapContainerMetrics extracts system metrics from m and appends them to all.
--- a/pkg/otlp/metrics/metrics_remapping_test.go
+++ b/pkg/otlp/metrics/metrics_remapping_test.go
@@ -275,15 +275,20 @@ func TestRemapMetrics(t *testing.T) {
 			out: []pmetric.Metric{metric("container.net.rcvd.packets", point{f: 15})},
 		},
 	} {
-		lena := dest.Len()
-		checkprefix := strings.HasPrefix(tt.in.Name(), "system.") || strings.HasPrefix(tt.in.Name(), "process.")
+		lenin := dest.Len()
 		remapMetrics(dest, tt.in)
-		if checkprefix {
-			require.True(t, strings.HasPrefix(tt.in.Name(), "otel."), "system.* and process.* metrics need to be prepended with the otel.* namespace")
+		out := tt.out
+		if strings.HasPrefix(tt.in.Name(), "system.") || strings.HasPrefix(tt.in.Name(), "process.") {
+			// on system.* and process.* metrics, we should expect an additional metric
+			// which is a copy of the input with the otel.* prefix added
+			cp := pmetric.NewMetric()
+			tt.in.CopyTo(cp)
+			cp.SetName("otel." + cp.Name())
+			out = append(out, cp)
 		}
-		require.Equal(t, dest.Len()-lena, len(tt.out), "unexpected number of metrics added")
-		for i, out := range tt.out {
-			require.Equal(t, out, dest.At(dest.Len()-len(tt.out)+i))
+		require.Equal(t, dest.Len()-lenin, len(out), "unexpected number of metrics added")
+		for i, o := range out {
+			require.Equal(t, o, dest.At(dest.Len()-len(out)+i))
 		}
 	}
 
--- a/.chloggen/gbbr_fix-otel-namespace.yaml
+++ b/.chloggen/gbbr_fix-otel-namespace.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component (e.g. pkg/quantile)
+component: pkg/otlp
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: System (and process.*) metrics were remapped with otel.*, overwriting the original metric and losing it. This change fixes that, keeping the original metric.
+
+# The PR related to this change
+issues: [141]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:


pkg/otlp/metrics/metrics_remapping.go
```diff
@@ -70,7 +70,9 @@ func remapSystemMetrics(all pmetric.MetricSlice, m pmetric.Metric) {
 		copyMetric(all, m, "system.disk.in_use", 1)
 	}
 	// process.* and system.* metrics need to be prepended with the otel.* namespace
-	m.SetName("otel." + m.Name())
+	newm := all.AppendEmpty()
```
songy23, line 73: Chatted with @liustanley , only the metrics from the previous `extractSystemMetrics` were kept: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/23445/files#diff-59a3a2cae4395f15cad4800806f9dc7680aed0f303dd896c0ca8663e0fd0ac35L56

For `process.*` metrics, only the ones with `otel.*` prefix should be exported, original ones are expected to be overriden.

songy23 (changes requested):

pkg/otlp/metrics/metrics_remapping.go
gbbr: Sorry but I don't see the difference between this implementation and the one in the linked code. All metrics that start with system.* and process.* seem to get the prefix. I don't see any special case for `process.*`

songy23: Both of system.* and process.* metrics get the otel. prefix - that's right. In addition to that, the original system.* metrics (no otel. prefix) should be preserved / copied, while process.* metrics  don't.

gbbr: I understand, but I don't see that in the old code. The behaviour seems the same for both process and system

gbbr: Either way, in my latest commit, I think that I have applied what you meant. PTAL

pkg/otlp/metrics/metrics_remapping.go
```diff
@@ -69,7 +69,12 @@ func remapSystemMetrics(all pmetric.MetricSlice, m pmetric.Metric) {
 	case "system.filesystem.utilization":
 		copyMetric(all, m, "system.disk.in_use", 1)
 	}
-	// process.* and system.* metrics need to be prepended with the otel.* namespace
+	if strings.HasPrefix(name, "system.") {
+		// we keep the original metric for system.* metrics
```
songy23, line 73: I looked again at the old code, I think the behavior is still different: we prepend `otel.` to the original names, but we keep the remapped names with no `otel.` prefix. https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/23445/files#diff-59a3a2cae4395f15cad4800806f9dc7680aed0f303dd896c0ca8663e0fd0ac35L115-L129

For example, in the old code:
- if we have `"system.filesystem.utilization"`, first we remap it to `"system.disk.in_use"` and append the new one to slice
- then we prepend `otel.` to `"system.filesystem.utilization"` while 
`"system.disk.in_use"` is left unchanged
- in the end we got `"otel.system.filesystem.utilization"` and `"system.disk.in_use"`

In short, we prepend `otel.` to the original metrics, but don't prepend it to the remapped metrics.

Sorry for the confusion, I was also trying to understand the old behavior.

gbbr: I think you are misunderstanding the current PR. It works exactly like you expect (in your "old code" example above). I have simplified the tests for you to make it easier for you to understand. Feel free to set up a call if it's still not clear.

pkg/otlp/metrics/metrics_remapping_test.go
```diff
@@ -55,22 +55,30 @@ func TestRemapMetrics(t *testing.T) {
...
 		in  pmetric.Metric
 		out []pmetric.Metric
 	}{
 		{
-			in:  metric("system.cpu.load_average.1m", point{f: 1}),
-			out: []pmetric.Metric{metric("system.load.1", point{f: 1})},
+			in: metric("system.cpu.load_average.1m", point{f: 1}),
+			out: []pmetric.Metric{
+				metric("system.load.1", point{f: 1}),
+				metric("system.cpu.load_average.1m", point{f: 1}),
```
dineshg13, line 66: shouldn't this be `otel.system.cpu.load_average.1m` ?

pkg/otlp/metrics/metrics_remapping_test.go
```diff
@@ -275,16 +322,20 @@ func TestRemapMetrics(t *testing.T) {
...
-		remapMetrics(dest, tt.in)
-		if checkprefix {
-			require.True(t, strings.HasPrefix(tt.in.Name(), "otel."), "system.* and process.* metrics need to be prepended with the otel.* namespace")
-		}
-		require.Equal(t, dest.Len()-lena, len(tt.out), "unexpected number of metrics added")
-		for i, out := range tt.out {
-			require.Equal(t, out, dest.At(dest.Len()-len(tt.out)+i))
-		}
+		t.Run("", func(t *testing.T) {
+			dest := pmetric.NewMetricSlice()
```
songy23, line 326: It would be more clear if `in` is in `dest` which is consistent with:
https://github.com/DataDog/opentelemetry-mapping-go/blob/3d07b09ccf1829bd6bbe804f65d5ee6756478776/pkg/otlp/metrics/metrics_translator.go#L661

https://github.com/DataDog/opentelemetry-mapping-go/blob/3d07b09ccf1829bd6bbe804f65d5ee6756478776/pkg/otlp/metrics/metrics_translator.go#L682

pkg/otlp/metrics/metrics_remapping.go
songy23: If I'm not mistaken: using the same example, now after `remapMetrics` we're getting 3 metrics: `"otel.system.filesystem.utilization"`, `"system.disk.in_use"` and `"system.filesystem.utilization"`?

If so, the last one `"system.filesystem.utilization"` shouldn't be there

pkg/otlp/metrics/metrics_remapping_test.go
```diff
@@ -155,11 +190,23 @@ func TestRemapMetrics(t *testing.T) {
...
+		{
+			in: metric("system.filesystem.utilization", point{f: 15}),
+			out: []pmetric.Metric{
+				metric("system.disk.in_use", point{f: 15}),
+				metric("system.filesystem.utilization", point{f: 15}),
 			},
 		},
 		{
-			in:  metric("system.filesystem.utilization", point{f: 15}),
-			out: []pmetric.Metric{metric("system.disk.in_use", point{f: 15})},
```
songy23, line 162: Hmmm I feel I'm getting more confused now - looks like the previous test is doing the the right thing, but how comes the resulting metric in datadog exporter got an `otel.` prefix? https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/23445#discussion_r1233776214

songy23:
I finally found the root cause - the bug isn't in this file `metrics_remapping.go`, but in `metrics_translator.go`: https://github.com/DataDog/opentelemetry-mapping-go/blob/3d07b09ccf1829bd6bbe804f65d5ee6756478776/pkg/otlp/metrics/metrics_translator.go#L660-L683

The for loop above is unfortunately an anti-pattern: you are appending more elements to `metricsArray` while iterating over it. For example, 

1. say initially `metricsArray` == [`system.filesystem.utilization`]
2. in the first iteration, you called `remapMetrics` so `system.disk.in_use` gets appended to `metricsArray`, and `system.filesystem.utilization` gets renamed to `otel.system.filesystem.utilization`. Now `metricsArray` == [`otel.system.filesystem.utilization`, `system.disk.in_use`]. If we stop here, everything is correct
3. however the outer for-loop will continue, so in the second iteration, you called `remapMetrics` again on `system.disk.in_use` so it gets prepended with `otel.` too. Now `metricsArray` == [`otel.system.filesystem.utilization`, `otel.system.disk.in_use`]. The second metrics is now incorrect.

Sent out https://github.com/DataDog/opentelemetry-mapping-go/pull/148 to fix.

PR #149, opened 2023-08-08
[pkg/otlp/metrics] Override sketch min and max with lowest and highest bounds

mx-psi:
### What does this PR do?

Overrides sketch min and max with the lower bound of the lowest nonzero bucket and the upper bound of the highest nonzero bucket respectively if these exist and are not infinite.

### Motivation

The quantile sketch estimation of minimum and maximum are an estimation that is sometimes wrong. This new way of estimating min/max ensures that the average is always within min and max.


songy23 (approved):
LGTM

PR #159, opened 2023-08-29
Remove go 1.19 support, bump minimum to go 1.20 and add testing for 1.21

liustanley:
### What does this PR do?

The previous golangci-lint version was built with go1.19 and was causing errors, so I upgraded it to version 1.54.2 which is built with go1.21.0. I'm not sure if we need to use an older version to use with go 1.20 though


internal/tools/go.mod
```diff
@@ -4,7 +4,7 @@ go 1.18
 
 require (
 	github.com/frapposelli/wwhrd v0.4.0
-	github.com/golangci/golangci-lint v1.53.3
+	github.com/golangci/golangci-lint v1.54.2
```
mx-psi, line 7: We use the same version on the Collector https://github.com/open-telemetry/opentelemetry-collector/blob/064df806f1caea2d95aaa407ff792e806f60896b/internal/tools/go.mod#L9 so this should be fine

mx-psi (approved):
You need to change the required jobs in the settings to be able to merge this

PR #171, opened 2023-10-02
[OTEL-948][pkg/otlp/metrics] Extract common interfaces to metricscommon

songy23:
### What does this PR do?

Extract and move common interfaces and functions from pkg/otlp/metrics to a new package pkg/otlp/metricscommon.

### Motivation

This is for resolving a dependency issue with gogo/protobuf. We will fork the pkg/otlp/metrics package to have no gogo/protobuf dependency and work with raw proto files instead.


songy23:
Realized this would also introduce breaking changes to pkg/otlp/metrics. Probably not worth doing the refactor to common.

PR #174, opened 2023-10-10
[pkg/inframetadata] Add support for host.cpu attributes

mx-psi:
### What does this PR do?

Adds support for upcoming CPU resource attributes defined by open-telemetry/semantic-conventions/pull/209.

### Motivation

OpenTelemetry customers have this information available for all hosts.

### DIFF
--- a/pkg/inframetadata/internal/hostmap/hostmap.go
+++ b/pkg/inframetadata/internal/hostmap/hostmap.go
@@ -43,10 +43,18 @@ func strField(m pcommon.Map, key string) (string, bool, error) {
 		// Field not available, don't update but don't fail either
 		return "", false, nil
 	}
-	if val.Type() != pcommon.ValueTypeStr {
+
+	var value string
+	switch val.Type() {
+	case pcommon.ValueTypeStr:
+		value = val.Str()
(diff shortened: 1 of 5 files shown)


pkg/inframetadata/internal/hostmap/hostmap.go
```diff
@@ -43,10 +43,18 @@ func strField(m pcommon.Map, key string) (string, bool, error) {
 		// Field not available, don't update but don't fail either
 		return "", false, nil
 	}
-	if val.Type() != pcommon.ValueTypeStr {
+
+	var value string
+	switch val.Type() {
+	case pcommon.ValueTypeStr:
+		value = val.Str()
+	case pcommon.ValueTypeInt:
```
gbbr, line 51: Maybe you should update the documentation (or name) of this function to indicate that it also returns an int field now.

gbbr (approved):

pkg/inframetadata/internal/hostmap/hostmap.go
mx-psi: Makes sense, fixed!

PR #184, opened 2023-10-30
[pkg/inframetadata] Add support for metrics for host metadata

mx-psi:
### What does this PR do?

Adds support for CPU and RAM metrics in host metadata:
- We take the latest point from the metric when updating metadata.
- As opposed to changes on resource attributes, we do not track updates to the host metadata payload and do not force sending a new host metadata payload, since for metrics like frequency this would lead to too frequent updates

### Motivation

These are commonly used fields in host metadata


pkg/inframetadata/internal/hostmap/hostmap.go
```diff
@@ -183,6 +187,48 @@ func (m *HostMap) Update(host string, res pcommon.Resource) (changed bool, md pa
...
 }
 
+func (m *HostMap) UpdateFromMetric(host string, metric pmetric.Metric) {
+	var point pmetric.NumberDataPoint
+
+	// Take last available point
+	switch metric.Type() {
+	case pmetric.MetricTypeGauge:
+		lastIndex := metric.Gauge().DataPoints().Len() - 1
+		point = metric.Gauge().DataPoints().At(lastIndex)
```
songy23, line 197: Is metric guaranteed to have at least 1 data point?

pkg/inframetadata/reporter.go
```diff
@@ -137,6 +147,36 @@ func (r *Reporter) ConsumeResource(res pcommon.Resource) error {
 	return nil
 }
 
+func (r *Reporter) ConsumeMetrics(md pmetric.Metrics) error {
```
songy23, line 150: Does this func need to be exported? If so considering adding docs

pkg/inframetadata/reporter.go
```diff
@@ -137,6 +147,36 @@ func (r *Reporter) ConsumeResource(res pcommon.Resource) error {
...
+	rms := md.ResourceMetrics()
+	for i := 0; i < rms.Len(); i++ {
+		rm := rms.At(i)
+		res := rm.Resource()
+		if ok, err := hasHostMetadata(res); err != nil {
+			return fmt.Errorf("failed to check resource: %w", err)
+		} else if !ok {
+			// The resource should not be used for host metadata.
+			return nil
```
songy23, line 159: Why return instead of continue?

pkg/inframetadata/internal/hostmap/hostmap.go
mx-psi: This does not seem to be a requirement on either the spec or opentelemetry-proto, so I added a check for it.

pkg/inframetadata/reporter.go
mx-psi: Yes, documented it! The idea is to call this on the exporter when consuming metrics

pkg/inframetadata/reporter.go
mx-psi: Oops, good catch, thanks!

songy23 (approved):
LGTM

PR #193, opened 2023-11-09
OTEL-738: Add Custom Container Tag Support (datadog.container.tag prefix)

mackjmr:
### What does this PR do?

This PR deprecates `ContainerTagFromAttributes` in favor of `ContainerTagFromResourceAttributes` which accepts a `pcommon.Map` and is now called for metrics, logs and traces. `ContainerTagFromResourceAttributes` not only checks semantic conventions for container tags, but also checks resource attributes which are prefixed by `datadog.container.tag.`.

Previously, `ContainerTagFromAttributes` was called by traces [here](https://github.com/DataDog/datadog-agent/blob/main/pkg/trace/api/otlp.go#L542). `ContainerTagFromResourceAttributes` accepting a `pcommon.Map` has a couple implications for tracing:
- We will now only read container tags from resource attributes, whereas we currently read them from both resource attributes and span attributes. Container Semantic Conventions are meant for resource attributes, not span attributes.
- We will stop adding the mapped container tags as span tags, but only add them in the container tag section (`_dd.tags.container`). `_dd.tags.container` is the intended place for container tags so they can be searched without a @, so there is no need to also set these as span tag (searchable with @).

### Motivation

[OTEL-738
](https://datadoghq.atlassian.net/browse/OTEL-738)


pkg/otlp/attributes/attributes.go
```diff
@@ -66,32 +71,8 @@ var (
...
-		conventions.AttributeCloudRegion,
-		conventions.AttributeCloudAvailabilityZone,
-		conventions.AttributeAWSECSTaskFamily,
-		conventions.AttributeAWSECSTaskARN,
-		conventions.AttributeAWSECSClusterARN,
-		conventions.AttributeAWSECSTaskRevision,
-		conventions.AttributeAWSECSContainerARN,
-	}
+	// customContainerTagPrefix defines the prefix for custom container tags.
+	customContainerTagPrefix = "datadog.container.tag."
```
mx-psi, line 75: nit: this one can be a constant instead of a variable

pkg/otlp/attributes/attributes.go
```diff
@@ -173,16 +160,39 @@ func OriginIDFromAttributes(attrs pcommon.Map) (originID string) {
 	return
 }
 
+// ContainerTagFromResourceAttributes extracts container tags from the given
```
mx-psi, line 163: IMO it would be nice to document here the `datadog.container.tag` prefix thing as part of the function's documentation.

Another thing that would be relevant to document is what happens in these two edge cases:
1. Two container tags with the same name but different value (`datadog.container.tag.foo`: `bar` and `datadog.container.tag.foo`: `baz`)
2. A container tag that has the same name as a DD container mapping (`datadog.container.tag.container_id`: `bar`, `container.id`: `baz`)

It could be that anything can happen, in any case we should explicitly state it.

mackjmr: For 1., I don't think that it is possible to have the same key twice (see [here](https://opentelemetry.io/docs/specs/otel/common/#attribute:~:text=The%20keys%20in%20each%20such%20collection%20are%20unique%2C%20i.e.%20there%20MUST%20NOT%20exist%20more%20than%20one%20key%2Dvalue%20pair%20with%20the%20same%20key.)). If a user sets the same key twice, this will be deduplicated (see [here](https://opentelemetry.io/docs/specs/otel/common/#attribute:~:text=Setting%20an%20attribute%20with%20the%20same%20key%20as%20an%20existing%20attribute%20SHOULD%20overwrite%20the%20existing%20attribute%E2%80%99s%20value)). So I think we can assume that each key is unique here ?

For 2. I think that the semantic convention should take priority, I will document this, add code for this as well as a test case.

For 3., any `pcommon.Value` where `.AsString()` returns an empty string, we will skip adding. I will document this and add a test case.

mackjmr: 2. and 3. should be addressed in the latest commit

mx-psi: > For 1., I don't think that it is possible to have the same key twice (see [here](https://opentelemetry.io/docs/specs/otel/common/#attribute:~:text=The%20keys%20in%20each%20such%20collection%20are%20unique%2C%20i.e.%20there%20MUST%20NOT%20exist%20more%20than%20one%20key%2Dvalue%20pair%20with%20the%20same%20key.)). If a user sets the same key twice, this will be deduplicated (see [here](https://opentelemetry.io/docs/specs/otel/common/#attribute:~:text=Setting%20an%20attribute%20with%20the%20same%20key%20as%20an%20existing%20attribute%20SHOULD%20overwrite%20the%20existing%20attribute%E2%80%99s%20value)). So I think we can assume that each key is unique here ?

You are right :) We can ignore this case. Incidentally, this is another 'feature' allowed in host tags but not here.

> For 2. I think that the semantic convention should take priority, I will document this, add code for this as well as a test case.

:+1:, agreed that that seems like the most reasonable approach.

> For 3., any pcommon.Value where .AsString() returns an empty string, we will skip adding. I will document this and add a test case.

Note that `AsString` will not return an empty string for any type; instead it will convert to a string with some predefined method (e.g. for a slice [it will marshal as JSON](https://github.com/open-telemetry/opentelemetry-collector/blob/2a9d870a57038cb228ef20e7620f18a93c92b9d3/pdata/pcommon/value.go#L398-L399)).

mackjmr: Apologies, I said `.AsString()` but meant [.Str()](https://pkg.go.dev/go.opentelemetry.io/collector/pdata@v1.0.0-rcv0017/pcommon#Value.Str). This is what we currently use for conventions mapping: https://github.com/DataDog/opentelemetry-mapping-go/blob/main/pkg/otlp/attributes/attributes.go#L147. 

To stay consistent, I think we should only add this for value type ValueTypeStr here as well. We can reconsider if there is ask for other value types.

mx-psi (approved):

PR #197, opened 2023-11-17
Add functionality for runtime metric mappings to sum values by attribute

liustanley:
### What does this PR do?

This PR refactors the runtime metrics mapping code and adds functionality to sum values by a specified attribute key. For example, the OTel metric `process.runtime.jvm.memory.usage` divides its values by different memory pools, which is designated by the `pool` attribute. In order to correctly map this metric to the DD metric `jvm.heap_memory`, we need to sum all the data point values with attribute `type: heap` and `pool:[any value]`. This applies to all of the JVM metrics that are divided by memory pools: 
- `process.runtime.jvm.memory.usage`
- `process.runtime.jvm.memory.committed`
- `process.runtime.jvm.memory.init`
- `process.runtime.jvm.memory.limit`.

### Motivation

- https://datadoghq.atlassian.net/browse/OTEL-1139
- https://datadoghq.atlassian.net/browse/OTELS-110

### DIFF
--- a/pkg/otlp/metrics/metrics_translator_test.go
+++ b/pkg/otlp/metrics/metrics_translator_test.go
@@ -607,7 +607,7 @@ func TestMapRuntimeMetricWithTwoAttributesHasMapping(t *testing.T) {
 		consumer.metrics,
 		[]metric{
 			newGaugeWithHost(newDims("process.runtime.jvm.memory.usage").AddTags("pool:G1 Old Gen", "type:heap"), uint64(seconds(startTs+1)), 10, fallbackHostname),
-			newGaugeWithHost(newDims("jvm.heap_memory").AddTags("pool:G1 Old Gen"), uint64(seconds(startTs+1)), 10, fallbackHostname),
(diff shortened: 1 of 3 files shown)


pkg/otlp/metrics/metrics_translator_test.go
```diff
@@ -607,7 +607,7 @@ func TestMapRuntimeMetricWithTwoAttributesHasMapping(t *testing.T) {
 		consumer.metrics,
 		[]metric{
```
mx-psi, line 608: Can we add, a test that has, in the same payload, metrics that differ in some user-set attribute but are equal in all other attributes?

mx-psi (commented):
General comment: Two downsides of this approach are the following:
1. It introduces the requirement that all metrics from the same JVM instance are sent to the same Datadog exporter instance. Is that a requirement we already had? 
2. It modifies the `pdata` payload, which is less efficient. I think we are already doing this, but still it would be good to avoid it if possible.

PR #203, opened 2023-11-27
[pkg/otlp/metrics] Add support for metrics origin

mx-psi:
### What does this PR do?

Adds support for setting metrics origin on metrics based on the scope.


mx-psi:
@mackjmr requesting your review since you reviewed #204, but this one is less urgent IMO than #231

pkg/otlp/metrics/origin.go
```diff
@@ -14,11 +14,198 @@
...
+	default:
+		return fmt.Sprintf("OriginCategory(%d)", o)
+	}
+}
+
 // OriginService defines the origin service.
 type OriginService int32
+
+// List all receivers that set the scope name.
+const (
```
mackjmr, line 66: Missing `jmxreceiver` and `podmanreceiver` which both produce metrics.

mackjmr: Ah nevermind, these two receivers don't set scope name.

mackjmr (approved):
A dedicated test in `metrics_translator_test.go` with something like:
```
	assert.ElementsMatch(t,
		consumer.metrics,
		[]metric{
			newCount(exampleDimsWithOrigin, uint64(seconds(1)), 30),
		},
	)

```

would be nice to have here, as this goes through the whole path. 

Other than that LGTM!

mx-psi:
> A dedicated test in `metrics_translator_test.go` with something like:
> 
> ```
> 	assert.ElementsMatch(t,
> 		consumer.metrics,
> 		[]metric{
> 			newCount(exampleDimsWithOrigin, uint64(seconds(1)), 30),
> 		},
> 	)
> ```
> 
> would be nice to have here, as this goes through the whole path.
> 
> Other than that LGTM!

All the tests using the files in `pkg/otlp/metrics/testdata` are already testing this. I don't have one specific for a metrics product, maybe I can add one of those?

pkg/otlp/metrics/origin.go
```diff
@@ -14,11 +14,198 @@
...
+
 // OriginProduct defines the origin product.
 type OriginProduct int32
 
+const (
+	OriginProductUnknown OriginProduct = 0
+	// OriginProductDatadogAgent is the origin for metrics coming from the Datadog Agent OTLP Ingest.
+	OriginProductDatadogAgent OriginProduct = 10
+	// OriginProductDatadogExporter is the origin for metrics coming from the OTel Collector Datadog Exporter.
+	OriginProductDatadogExporter OriginProduct = 17
```
mx-psi, line 30: I am going to mark back as draft until we confirm this

mackjmr:
> All the tests using the files in pkg/otlp/metrics/testdata are already testing this. I don't have one specific for a metrics product, maybe I can add one of those?

Yeah, I think this would be good. Also, not sure if it is possible, but if we can set the scope in one of the tests it would be nice to include a known origin service in there.

pkg/otlp/metrics/origin.go
```diff
@@ -14,11 +14,198 @@
...
+		return "otlp"
+	default:
+		return fmt.Sprintf("OriginCategory(%d)", o)
+	}
+}
+
 // OriginService defines the origin service.
 type OriginService int32
+
+// List all receivers that set the scope name.
```
mackjmr, line 65: I think this should also contain `connectors` ? A few add the scope: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/connector/countconnector/connector.go#L79.

pkg/otlp/metrics/origin.go
mx-psi: Removing it for now

pkg/otlp/metrics/origin.go
mx-psi: Spoke offline about this, makes sense but will add it later since it needs to go through backend approval

mx-psi:
Added a specific test for this, auto-merging....

PR #217, opened 2023-12-11
[POC][DO NOT MERGE] Kafka

mackjmr:


pkg/otlp/metrics/metrics_translator.go
```diff
@@ -40,7 +40,21 @@ const (
 	errNoBucketsNoSumCount string = "no buckets mode and no send count sum are incompatible"
 )
 
-var _ source.Provider = (*noSourceProvider)(nil)
+var (
+	_ source.Provider = (*noSourceProvider)(nil)
+
+	rateAsGaugeMetrics = map[string]string{
```
mx-psi, line 46: Typically if what you want to represent is a set, you can use a `map[string]struct{}` and check 

```go
if _, ok := rateAsGaugeMetrics[dims.name]; ok {
   // do something
}

```

pkg/otlp/metrics/ttlcache.go
```diff
@@ -82,6 +82,13 @@ func (t *ttlCache) putAndGetDiff(
 			return 0, false
 		}
 		dx = val - cnt.value
+
+		// For the purpose of the POC using a map here for simplicity. Will likely 
+		// need to create a new func to use in this case.
+		if ok := rateAsGaugeMetrics[dimensions.name]; ok == "ok" {
+			dx = dx / time.Duration(ts - cnt.ts).Seconds()
```
mx-psi, line 89: (For the refactor) We should check that `dt` is not zero, I think it makes sense to just change the `if cnt.ts > ts` above

mx-psi (commented):
The approach makes sense to me

pkg/otlp/metrics/ttlcache.go
mackjmr: > We should check that dt is not zero

Not sure to understand the suggestion, what is `dt` ?

mx-psi: Sorry, I meant the difference between `ts` and `cnt.ts`, the rate is the ratio between the delta in values and the delta in time, which you usually represent as $\mathrm{d}t$: 

$$\mathrm{rate} = \frac{\mathrm{d}x}{\mathrm{d}t} = \frac{x_{\mathrm{now}} - x_{\mathrm{prev}}}{t_{\mathrm{now}} - t_{\mathrm{prev}}}$$

mackjmr: Understood, thanks for clarifying 👍

PR #220, opened 2023-12-12
[pkg/otlp/metrics] Add telemetry metric counting the number of missing sources

mx-psi:
### What does this PR do?

Creates a new metric, `datadog.otlp_translator.metrics.missing_source` that counts the number of metrics that are missing a source.

### Motivation

We want users to be able to control how many metrics are being tagged automatically by the exporter/Agent.


pkg/otlp/metrics/metrics_translator.go
```diff
@@ -51,9 +55,10 @@ func (*noSourceProvider) Source(context.Context) (source.Source, error) {
 
 // Translator is a metrics translator.
 type Translator struct {
-	prevPts *ttlCache
-	logger  *zap.Logger
-	cfg     translatorConfig
+	prevPts        *ttlCache
+	logger         *zap.Logger
+	missingSources otelmetric.Int64Counter
```
mackjmr, line 60: I'm not sure how many more metrics we will add, but my concern with adding this as a field in the `Translator` struct is then each time we want to add a new metric we need to add a field to this struct.

It may be easier to have a separate struct that stores all our instruments:
```
type Translator struct {
	instruments instruments
}

type instruments struct {
	missingSources otelmetric.Int64Counter
	othermetric    otelmetric.Float64Histogram
}
```

What do you think ?

mx-psi: Makes sense, changed it!

mackjmr (approved):

PR #222, opened 2023-12-14
POC for trsnaltor for attributes

mx-psi:


pkg/otlp/attributes/attributes_translator.go
```diff
@@ -0,0 +1,61 @@
...
+		return nil, fmt.Errorf("failed to build missing source counter: %w", err)
+	}
+
+	return &Translator{
+		missingSources: missingSources,
+	}, nil
+}
+
+// MapToSource gets a telemetry signal source from its attributes.
+func (p *Translator) MapToSource(ctx context.Context, attrs pcommon.Map, set attribute.Set) (source.Source, bool) {
```
mackjmr, line 54: Seems like the PR is missing the logic to avoid counting `datadog.otlp_translator.attributes.missing_source` ?

pkg/otlp/logs/logs_translator.go
```diff
@@ -194,13 +212,15 @@ func flattenAttribute(key string, val pcommon.Value, depth int) map[string]strin
 	return result
 }
 
-func extractHostNameAndServiceName(resourceAttrs pcommon.Map, logAttrs pcommon.Map) (host string, service string) {
-	if src, ok := attributes.SourceFromAttrs(resourceAttrs); ok && src.Kind == source.HostnameKind {
+func (t *Translator) extractHostNameAndServiceName(resourceAttrs pcommon.Map, logAttrs pcommon.Map) (host string, service string) {
+	// TODO: This maps the same resource multiple times and generates the metric multiple times :(
```
mackjmr, line 216: This should be fixable by refactoring logs code and taking `plog.ResourceLogsSlice` [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/datadogexporter/logs_exporter.go#L106) and [here](https://github.com/DataDog/datadog-agent/blob/main/comp/otelcol/otlp/internal/logsagentexporter/logs_exporter.go#L41) as an argument, but yeah don't think this is in the scope of the work here

pkg/otlp/attributes/attributes_translator.go
```diff
@@ -0,0 +1,61 @@
...
+
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/pdata/pcommon"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
+
+	"github.com/DataDog/opentelemetry-mapping-go/pkg/otlp/attributes/source"
+)
+
+const missingSourceMetricName string = "datadog.otlp_translator.attributes.missing_source"
```
mackjmr, line 29:
```suggestion
const missingSourceMetricName string = "datadog.otlp_translator.resources.missing_source"
```

Technically we are counting the number of resources without host identifying attributes, rather than attributes ?

pkg/otlp/attributes/attributes_translator.go
```diff
@@ -0,0 +1,61 @@
...
+	missingSources metric.Int64Counter
+}
+
+// NewTranslator returns a new attributes mapper.
+func NewTranslator(set component.TelemetrySettings) (*Translator, error) {
+	meter := set.MeterProvider.Meter("github.com/DataDog/opentelemetry-mapping-go/pkg/otlp/attributes")
+	missingSources, err := meter.Int64Counter(
+		missingSourceMetricName,
+		metric.WithDescription("OTLP payloads that are missing a source (e.g. hostname)"),
+		metric.WithUnit("[payload]"),
```
mackjmr, line 42: may need to change this to `{resource}` if you agree with the suggestion above. In both cases I believe it needs to be [curly braces](https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-units).

mackjmr (commented):
This approach looks good to me 👍 Only _important_  part missing is to avoid counting the metric itself, other than that just a few nits.

pkg/otlp/attributes/attributes_translator.go
mackjmr: (Assuming the logs code gets refactored)

mx-psi:
Last part in #231, where I addressed all the comments here :)

PR #223, opened 2023-12-14
[pkg/otlp/logs] Ignore resource attributes semantic conventions outside of the resource

mx-psi:
### What does this PR do?

Ignores `service.name` and hostname-related resource attributes outside of the resource level for logs.

### Motivation

These semantic conventions are for resource attributes only, recognizing them at the log level is not in the specification and thus can be argued to be incorrect.


mx-psi:
Closing until we have a migration plan for this

PR #228, opened 2023-12-19
[OTEL-1325] [Monotonic] Drop points with timestamp older or equal to timestamp of previous point received

mackjmr:
### What does this PR do?

When we receive two subsequent values with the same timestamp, the second value received will be ignored for diffs. Currently we still compute the diff.

This PR also fixes an issue where a point could be considered and computed as the first value, if it's timestamp was older than the timestamp of the point we currently have in the cache. Now, the first point calculation will not be considered for points which are older or equal to the current point in cache (having a current point in cache indicates that it is not the first point).

This is the First PR to compute rates. In the case of rates, having the same timestamp would mean that `dt` == 0 so we cannot compute the rate.


pkg/otlp/metrics/metrics_translator.go
```diff
@@ -216,7 +216,7 @@ func (t *Translator) mapNumberMonotonicMetrics(
 
 		if dx, ok := t.prevPts.MonotonicDiff(pointDims, startTs, ts, val); ok {
 			consumer.ConsumeTimeSeries(ctx, pointDims, Count, ts, dx)
-		} else if i == 0 && t.shouldConsumeInitialValue(startTs, ts) {
+		} else if i == 0 && dx >= 0 && t.shouldConsumeInitialValue(startTs, ts) {
```
mx-psi, line 219: Not sure I get this, if `dx == 0`, then we know the point in the cache is more recent, so maybe we should not report this first point?

mackjmr: If there is no point for the metric in the cache, [this code](https://github.com/DataDog/opentelemetry-mapping-go/blob/8a7737db71b42ace42d41b490fe158e7e168a799/pkg/otlp/metrics/ttlcache.go#L77-L88) won't run and `dx` will be zero. In this case, we need to compute the first point as this point not being in the cache indicate this is likely the first value.

I changed the return [here](https://github.com/DataDog/opentelemetry-mapping-go/pull/228/files#diff-0fb69cae87533a77406de810cfa4b0233f043eb777668b55f43cf46623796325R82) to `-1` so we know that the value is not the first point if this is negative.

songy23 (commented):
Needs `make fmt`

mx-psi (approved):

mackjmr:
@mx-psi This PR is now ready for review again

pkg/otlp/metrics/metrics_translator_test.go
```diff
@@ -371,6 +373,221 @@ func TestMapIntMonotonicWithReboot(t *testing.T) {
...
+		assert.ElementsMatch(t,
+			consumer.metrics,
+			[]metric{
+				newCountWithHost(exampleDims, uint64(seconds(startTs+4)), 30, fallbackHostname),
+			},
+		)
+		assert.Empty(t, rmt.Languages)
+	})
+
+	t.Run("older", func(t *testing.T) {
```
mackjmr, line 550: This test fails on `main` without the changes in this PR. With the old logic, an old point could be computed as a first val erroneously.

pkg/otlp/metrics/metrics_translator.go
```diff
@@ -202,9 +202,17 @@ func (t *Translator) mapNumberMonotonicMetrics(
 			continue
 		}
 
-		if dx, ok := t.prevPts.MonotonicDiff(pointDims, startTs, ts, val); ok {
+		dx, firstPoint, dropPoint := t.prevPts.MonotonicDiff(pointDims, startTs, ts, val)
```
mx-psi, line 205: nit: I think being a bit more verbose here helps readability
```suggestion
		dx, isFirstPoint, shouldDropPoint := t.prevPts.MonotonicDiff(pointDims, startTs, ts, val)
```

pkg/otlp/metrics/metrics_translator.go
```diff
@@ -202,9 +202,17 @@ func (t *Translator) mapNumberMonotonicMetrics(
...
+		if dropPoint {
+			t.logger.Debug("Dropping point: timestamp is older or equal to timestamp of previous point received", zap.String(metricName, pointDims.name))
+			continue
+		}
+
+		if !firstPoint {
 			consumer.ConsumeTimeSeries(ctx, pointDims, Count, ts, dx)
 		} else if i == 0 && t.shouldConsumeInitialValue(startTs, ts) {
+			// We only compute the first point in the timeseries if it is the first value in the datapoint slice.
+			// Todo: Investigate why we don't compute first val if i > 0 and add reason as comment.
```
mx-psi, line 215: @songy23 We may have discussed this when you aded this functionality, but I don't remember. Do you recall the rationale behind this?

songy23: We only consume the initial value for cumulative metrics. We do so by calculating the delta as if the previous value is 0. When `i>0` we just calculate the real delta (`pts[i] - pts[i-1]`)

mackjmr: But if there is a reset within a datapoint slice, e.g. values of `[]int64{10, 30, 5, 20}`.

In this case, there is a reset at index 2 and we can't compute the diff because it would be negative. Because of the `i>0`, we don't compute 5 as the initial value, but only put the point in the cache. Is this the intended behaviour ? Or should we consume 5 as the initial value here ?

mackjmr: These two tests: https://github.com/DataDog/opentelemetry-mapping-go/blob/de6046116470f8e2367b9a38213a7bafe53eee0f/pkg/otlp/metrics/metrics_translator_test.go#L350-L417 show how it currently is.

**TestMapIntMonotonicWithRebootWithinSlice**
- datapoint slice 1: `[]int64{0, 30, 0, 20}`

**TestMapIntMonotonicWithRebootBeginningOfSlice**
- datapoint slice 1: `[]int64{10}`
- datapoint slice 2: `[]int64{5, 30}`

You can see that in the first test, the value at reset is not consumed as first value because it is within the slice, whereas in the second test the value at reset is consumed as first value because it is at index 0 of the slice.

songy23: > But if there is a reset within a datapoint slice, e.g. values of []int64{10, 30, 5, 20}.

That's a different case. It is not handled in this `else if` block (it may be caught by the cache calculation but I'm not 100% sure). This block `else if i == 0 && t.shouldConsumeInitialValue(startTs, ts)` only takes care of the first value in the slice.

mackjmr: > it may be caught by the cache calculation

It is caught by the cache calculation (`dx < 0`), but the point does not get computed as an initial value, it only get's cached. This is not due to the changes in this PR, as there was the same behaviour before.

I'll create a task to investigate whether this is a feasible situation (are there ever resets within a slice), and then to update the code based on that. If it's a feasible scenario, then we should compute as first val, if it's not a feasable scenario then we should make a note in the code that this should never happen.

I don't think this should block this PR as there was the same behaviour before.

pkg/otlp/metrics/metrics_translator.go
```diff
@@ -202,9 +202,17 @@ func (t *Translator) mapNumberMonotonicMetrics(
 			continue
 		}
 
-		if dx, ok := t.prevPts.MonotonicDiff(pointDims, startTs, ts, val); ok {
+		dx, isFirstPoint, shouldDropPoint := t.prevPts.MonotonicDiff(pointDims, startTs, ts, val)
+		if shouldDropPoint {
+			t.logger.Debug("Dropping point: timestamp is older or equal to timestamp of previous point received", zap.String(metricName, pointDims.name))
```
mx-psi, line 207: Were we logging this before? I think we should avoid logging this, it's in the hot path and the volume of metrics processed by this could be huge. I like the explanation on https://github.com/open-telemetry/opentelemetry-collector/blob/main/CONTRIBUTING.md#logging:

> Use logging carefully for events that can happen frequently to avoid flooding the logs. Avoid outputting logs per a received or processed data item since this can amount to very large number of log entries (Collector is designed to process many thousands of spans and metrics per second). For such high-frequency events instead of logging consider adding an internal metric and increment it when the event happens.

mackjmr: We weren't logging it before -- Agreed it is likely this will spam the logs, but it will only get logged if it's in `debug` mode. Do you think this is enough of a safeguard, or I should still remove it ?

mx-psi: We do have other logs for other unlikely situations below... https://github.com/DataDog/opentelemetry-mapping-go/blame/de6046116470f8e2367b9a38213a7bafe53eee0f/pkg/otlp/metrics/metrics_translator.go#L780-L812 I guess we can keep it and think about switching to metrics/improving this

PR #230, opened 2023-12-19
[pkg/otlp/logs] Switch to Translator design

mx-psi:
### What does this PR do?

Move to a `Translator` struct design for logs. This is part of the refactor for accomplishing #222, I need to do this so that:
1. We can hold an attributes translator that has a counter inside the `logs.Translator`
2. It's easier for me to refactor extracting the hostname from the resource. Right now we call `SourceFromAttrs` once per log record, we can just call it once per resource.


mx-psi:
The usage on the Agent would look like this https://github.com/DataDog/datadog-agent/blob/bca34b6e8668b8986486e74e2e88c33be81d34b1/comp/otelcol/otlp/internal/logsagentexporter/logs_exporter.go#L41-L68

dineshg13 (approved):
LGTM - can merge after fixing the file name

PR #233, opened 2023-12-22
[pkg/inframetadata] Add support for host tags

mx-psi:
### What does this PR do?

- Resource attributes prefixed by `datadog.host.tag.` are now added as host tags for the host associated with the resource.
- Resource attributes for deployment environment and cluster name are now automatically added as host tags for the host associated with the resource.

Edge cases:

- If the type of a `datadog.host.tag.` prefixed attribute is not string, en error is raised
- If a `datadog.host.tag.` attribute has an empty string as a value, an error is raised
- Host tags are not updated per-tag but rather as a whole. For example, if a payload has `datadog.host.tag.foo -> bar` and `datadog.host.tag.baz -> qux` as resource attributes, and the next one only has `datadog.host.tag.foo -> bar`, the final payload after these transformations will only have `foo:bar` and **NOT** `baz:qux`. This is different from other attributes, but allows for removing host tags.

### Motivation

Allow users to add host tags for remote hosts.

### DIFF
--- a/.chloggen/host-tags-auto.yaml
+++ b/.chloggen/host-tags-auto.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component (e.g. pkg/quantile)
+component: pkg/inframetadata
+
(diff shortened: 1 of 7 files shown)


mackjmr (approved):
LGTM

PR #238, opened 2024-01-03
[pkg/otlp/metrics] Add support for computing rates

mackjmr:
### What does this PR do?
This PR adds support for rate as gauge metrics. For any monotonic metric present in the rateAsGauge map, we will compute the rate and submit the metric as a gauge. First val for rateAsGauge metrics are cached but not submitted as first val.

### Motivation

This PR is necessary to add support for kafka metrics that compute a rate and are submitted as a gauge.


pkg/otlp/metrics/metrics_translator.go
```diff
@@ -43,7 +43,10 @@ const (
 )
 
 var (
-	signalTypeSet = attribute.NewSet(attribute.String("signal", "metrics"))
+	signalTypeSet      = attribute.NewSet(attribute.String("signal", "metrics"))
+	rateAsGaugeMetrics = map[string]struct{}{
+		"placeholder.rate.as.gauge.metric": {},
```
mx-psi, line 48: You could override this map in tests and avoid having this in actual code (if different tests expected different values on the map, then moving this inside the translator may make sense, but I think it's not necessary right now AFAICT)

pkg/otlp/metrics/metrics_translator_test.go
```diff
@@ -268,37 +268,73 @@ func seconds(i int) pcommon.Timestamp {
...
-	// Create list of values
-	deltas := []int64{1, 2, 200, 3, 7, 0}
-	cumulative := make([]int64, len(deltas)+1)
-	cumulative[0] = 0
-	for i := 1; i < len(cumulative); i++ {
-		cumulative[i] = cumulative[i-1] + deltas[i-1]
-	}
+	t.Run("diff", func(t *testing.T) {
+		// Create list of values
+		deltas := []int64{1, 2, 200, 3, 7, 0}
```
mx-psi, line 276: You can reuse the code defining the point for both subtests (create a helper function to build a new point for you)

pkg/otlp/metrics/metrics_translator_test.go
```diff
@@ -350,70 +386,139 @@ func TestMapIntMonotonicDifferentDimensions(t *testing.T) {
 // This test checks that in the case of a reboot within a NumberDataPointSlice,
 // we cache the value but we do NOT compute first value for the value at reset.
 func TestMapIntMonotonicWithRebootWithinSlice(t *testing.T) {
-	values := []int64{0, 30, 0, 20}
-	slice := pmetric.NewNumberDataPointSlice()
-	slice.EnsureCapacity(len(values))
+	t.Run("diff", func(t *testing.T) {
+		values := []int64{0, 30, 0, 20}
```
mx-psi, line 390: Same here, you can just have a general helper instead of repeating the code

pkg/otlp/metrics/ttlcache_test.go
```diff
@@ -28,29 +29,57 @@ func newTestCache() *ttlCache {
...
+		assert.True(t, dropPoint, "new ts < old ts")
+
+		_, firstPoint, dropPoint = prevPts.MonotonicDiff(dims, startTs, 2, 2)
+		assert.True(t, firstPoint, "new < old => there has been a reset")
+		assert.False(t, dropPoint, "first point")
+
+		dx, firstPoint, dropPoint := prevPts.MonotonicDiff(dims, startTs, 3, 4)
+		assert.False(t, firstPoint)
+		assert.False(t, dropPoint)
+		assert.Equal(t, 2.0, dx, "expected diff 2.0 with (0,2,2) value")
```
mx-psi, line 55: I realize this was 'wrong' before your PR, but since you are adding new tests I am going to suggest we rewrite this as something like the following:

```go
// May not compile as-is, just wrote this without checking

points  := []struct{
  ts uint64
  val float64
  expectFirstPoint bool
  expectDropPoint bool
  dropPointMessage string
}{
// ...
}

var dx float64
startTs := uint64(0) // equivalent to start being unset
prevPts := newTestCache()
for _, point := range points {
  dx, firstPoint, dropPoint := prevPts.MonotonicDiff(dims, startTs, point.ts, point,val)
  assert.Equal(t, point.expectFirstPoint, firstPoint)
  assert.Equal(t, point.expectDropPoint, dropPoint, point.dropPointMessage)
}

assert.Equal(t, 2.0, dx, "...")
```

mx-psi (commented):
The logic looks good to me, but I am going to suggest we refactor the tests a bit so that it is more manageable.

edit: ~~I think that we have discussed this before, but note that rates usually have an interval at Datadog, and OTLP metrics don't have an interval, so it's kind of difficult to do the exact mapping some other Datadog products would do, and this may result in differences in-app. We can try and fix this later in any case.~~ This is not the case for these rate metrics

mackjmr:
> I think that we have discussed this before, but note that rates usually have an interval at Datadog, and OTLP metrics don't have an interval, so it's kind of difficult to do the exact mapping some other Datadog products would do, and this may result in differences in-app. We can try and fix this later in any case.

This PR doesn't add support for sending metrics as rates, but computing the rate and submitting it as a gauge.

I think what you mentioned is for actual in-app rates ?

If so, because this metric gets submitted as a gauge (that represents a rate), this shouldn't be an issue as the DD kafka metrics that I am trying to map are also in-app gauges that represents rates, so the mapping should be exact here.

mx-psi:
Yeah, you are completely right, sorry, the holidays weren't good for my memory :smile:

pkg/otlp/metrics/metrics_translator_test.go
mackjmr: Good callout, I added a helper. Let me know if it looks good ?

pkg/otlp/metrics/ttlcache_test.go
mackjmr: Done. I also tried to do the same with `TestMonotonicDiffKnownStart`, but needed to split into multiple slices for the resets.

pkg/otlp/metrics/metrics_translator.go
mackjmr: Done, I've overriden `rateAsGaugeMetrics` in tests and left it empty in `metrics_translator.go`. Once I've filled `rateAsGaugeMetrics` in a subsequent PR, I think I will be able to remove the overrides in test and just update [here](https://github.com/DataDog/opentelemetry-mapping-go/blob/aabd94ecc082197c4c7ff94c5f07cb335feb111f/pkg/otlp/metrics/metrics_translator_test.go#L271) with one of the "real" metrics, as this is unlikely to change.

mx-psi (approved):

pkg/otlp/metrics/metrics_translator.go
mx-psi: Makes sense, thanks!

pkg/otlp/metrics/ttlcache_test.go
mx-psi: Awesome, looks much better IMO, thanks :)

PR #239, opened 2024-01-05
[OTEL-1342][pkg/otlp/metrics] Add kafka metrics mappings

mackjmr:
### What does this PR do?

This PR adds the mappings for kafka metrics.


pkg/otlp/metrics/metrics_remapping.go
```diff
@@ -42,32 +44,32 @@ func remapSystemMetrics(all pmetric.MetricSlice, m pmetric.Metric) {
 	}
 	switch name {
 	case "system.cpu.load_average.1m":
-		copyMetric(all, m, "system.load.1", 1)
+		copyMetricWithAttr(all, m, "system.load.1", 1, attributesMapping{})
```
mx-psi, line 47: nit: I think we should make `attributesMapping{}` a package level constant to avoid possibly allocating a new one for each of these calls

pkg/otlp/metrics/metrics_remapping.go
```diff
@@ -165,6 +182,19 @@ func copyMetric(dest pmetric.MetricSlice, m pmetric.Metric, newname string, div
...
+				dp.Attributes().PutStr(k, v)
+			}
+		}
+		if dynamic := attributesMapping.dynamic; dynamic != nil {
+			for old, new := range dynamic {
+				if v, ok := dp.Attributes().Get(old); ok {
+					dp.Attributes().PutStr(new, v.AsString())
+				}
+			}
+		}
```
mx-psi, line 197: nit: No need for a `nil` check here: https://go.dev/play/p/2VYK2B78o8R

pkg/otlp/metrics/metrics_remapping.go
```diff
@@ -165,6 +182,19 @@ func copyMetric(dest pmetric.MetricSlice, m pmetric.Metric, newname string, div
...
+		// attributes mapping
+		if fixed := attributesMapping.fixed; fixed != nil {
+			for k, v := range fixed {
+				dp.Attributes().PutStr(k, v)
+			}
+		}
+		if dynamic := attributesMapping.dynamic; dynamic != nil {
+			for old, new := range dynamic {
+				if v, ok := dp.Attributes().Get(old); ok {
+					dp.Attributes().PutStr(new, v.AsString())
```
mx-psi, line 194: nit: We can delay mapping to string and instead do this to preserve the type:
```suggestion
					v.CopyTo(dp.Attributes().PutEmpty(new))
```

pkg/otlp/metrics/metrics_remapping.go
mackjmr: Good call, created `var emptyAttributesMapping = attributesMapping{} `

pkg/otlp/metrics/metrics_remapping.go
mackjmr: Done 👍

mx-psi (approved):

PR #248, opened 2024-01-17
[pkg/inframetadata] Use map[string]any everywhere

mx-psi:
### What does this PR do?

Align Gohai types with those returned by `github.com/DataDog/gohai`.

### Motivation

This could lead to a panic when mixing the two.


mackjmr (approved):
Approved although it would be good to have a regression test that panics without the fix in this PR.

mx-psi:
> Approved although it would be good to have a regression test that panics without the fix in this PR.

Agreed, will have to add that to the Datadog exporter instead of here, it is specific to what we do there

PR #250, opened 2024-01-19
[pkg/inframetadata] More robust gohai

mx-psi:
### What does this PR do?

- Gives concrete types to all Gohai fields.
- Change return type for `Network` method to `map[string]any`

### Motivation

Avoid panics because of casting to incorrect types.


pkg/inframetadata/gohai/internal/gohaitest/gohai_test.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/DataDog/gohai/cpu"
+	"github.com/DataDog/gohai/filesystem"
+	"github.com/DataDog/gohai/memory"
+	"github.com/DataDog/gohai/network"
+	"github.com/DataDog/gohai/platform"
+	"github.com/DataDog/opentelemetry-mapping-go/pkg/inframetadata/gohai"
+	"github.com/stretchr/testify/require"
+)
+
+func TestGohaiHasTheRightTypes(t *testing.T) {
```
mx-psi, line 18: Added a test here using gohai. This copies some code from the exporter, but given how likely it is (or, at least, I am) to cause trouble because of types, I am copying it here in a separate module to avoid pulling `Datadog/gohai` unnecessarily. Will also make some changes on the exporter to further mitigate issues here.

mackjmr (approved):

PR #262, opened 2024-01-30
[pkg/otlp/metrics][OTEL-1652] Support empty histograms

### LINKED ISSUE #252: OTEL Histograms without buckets are not exported
## Problem Statement:
When the OTEL Histogram is sent to DD Exporter without buckets, it seems the datapoints are never sent to the back-end. 

## Hypothesis: I guess the issue is with [getSketchBuckets](https://github.com/DataDog/opentelemetry-mapping-go/blob/81d9650bebbbcff44cec39ac97fd9c46420c32bc/pkg/otlp/metrics/metrics_translator.go#L264) which should produce a consumable sketch. Since there are no buckets in the histogram datapoint, [this loop](https://github.com/DataDog/opentelemetry-mapping-go/blob/81d9650bebbbcff44cec39ac97fd9c46420c32bc/pkg/otlp/metrics/metrics_translator.go#L283-L321) never get executed and sketch is always empty ([this condition is true](https://github.com/DataDog/opentelemetry-mapping-go/blob/81d9650bebbbcff44cec39ac97fd9c46420c32bc/pkg/otlp/metrics/metrics_translator.go#L324)) and hence the sketch is not submitted to the consumer.

The below test further proved my hypothesis but I would still leave it to whomever works on this issue. I had posted a Histogram with only one bucket count and no bounds and I see that getting exported with min bound as -Inf.

<details>
 
<summary>OTEL Histogram with just bucket count</summary>

```

Metric #0
Descriptor:
     -> Name: test.histogram
     -> Description: 
     -> Unit: seconds
     -> DataType: Histogram
     -> AggregationTemporality: Delta
HistogramDataPoints #0
StartTimestamp: 2024-01-19 14:24:35 +0000 UTC
Timestamp: 2024-01-19 14:24:40 +0000 UTC
Count: 10
Sum: 5.124000
Max: 0.936000
Buckets #0, Count: 10

```

</details>

<details>
 
<summary>DD Exported Sketch ([logged here](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/datadogexporter/metrics_exporter.go#L253))</summary>

```
{
  "metric": "test.histogram",
  "interval": 1,
  "points": [
    {
      "sketch": {
        "summary": {
          "Min": -9.941854089121368e-10,
          "Max": 0.936,
          "Sum": 5.124,
          "Avg": 0.5124,
          "Cnt": 10
        }
      },
      "ts": 1705674280
    }
  ]
}

```

</details>

As soon as that single bucket is removed, the above hypothesis seems to happen and the sketch is ignored. In this scenario the above mentioned log is never called as the skecth was not consumed. To further support this, when `send_aggregations` is enabled, this [block is executed ](https://github.com/DataDog/opentelemetry-mapping-go/blob/81d9650bebbbcff44cec39ac97fd9c46420c32bc/pkg/otlp/metrics/metrics_translator.go#L451-L469) and I could see `count`, `sum`, `max` getting reported as separate time-series.

## Expected Behaviour:
When a histogram with no buckets is exported, DD Distribution should still be exported and available for querying.

mx-psi:
### What does this PR do?

- [pkg/otlp/metrics] Refactor histogram loop to use slice
- [pkg/otlp/metrics] Add a single bucket for empty histograms

### Motivation

Fixes #252 by adding buckets representing the information conveyed by min, max, and count in the case a histogram is empty.

### DIFF
--- a/.chloggen/mx-psi_empty-histograms.yaml
+++ b/.chloggen/mx-psi_empty-histograms.yaml
@@ -1,0 +1,17 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component (e.g. pkg/quantile)
+component: pkg/otlp/metric
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Fixes a bug where empty histograms were not being sent to the backend in the distributions mode.
+
+# The PR related to this change
+issues: [252]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: | 
+  - Empty histograms are now mapped as if they had a single (min, max) bucket.
--- a/pkg/otlp/metrics/histograms_test.go
+++ b/pkg/otlp/metrics/histograms_test.go
@@ -77,6 +77,38 @@ func TestDeltaHistogramTranslatorOptions(t *testing.T) {
 			},
 		},
 		{
+			name:     "empty-delta-no-min-max",
+			otlpfile: "testdata/otlpdata/histogram/empty-delta-no-min-max.json",
+			ddogfile: "testdata/datadogdata/histogram/empty-delta-no-min-max.json",
+			options: []TranslatorOption{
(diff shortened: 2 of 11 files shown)


mackjmr (approved):
LGTM

PR #305, opened 2024-03-18
[pkg/otlp/attributes] Add warning when a resource is missing a hostname-identifying attribute

mx-psi:


songy23 (approved):

mx-psi (commented):
I am going to close this, I don't think this is the right place to put this in, since it would also log on the Datadog Agent's OTLP ingest. 

I guess the right place to put this is on the fallback hostname resolver in the exporter, but to do this I first have to avoid getting a hostname for initializing the trace agent

PR #320, opened 2024-04-19
[OTEL-1738] Deprecate APMStatsConsumer

songy23:
`APMStatsConsumer` is no longer in use since https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/31416. 

The only breaking change is users can no longer do `consumer.(*Consumer).ConsumeAPMStats()`, instead they can do `consumer.(*APMStatsConsumer).ConsumeAPMStats()`. Everything else is backwards-compatible.


.chloggen/deprecate-apm-stats-consumer.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: deprecation
```
mx-psi, line 2: Technically a breaking change? This changelog is mostly for us tho, so I am fine with either

mx-psi (approved):

.chloggen/deprecate-apm-stats-consumer.yaml
songy23: That makes sense, I updated the changelog

Issue #329, opened 2024-05-10
Feature Request: Add option to map array attributes into multiple tags with the same key


squah-confluent:
[OpenTelemetry supports array-valued attributes](https://github.com/open-telemetry/opentelemetry-specification/issues/341). When a metric with an array-valued attribute is exported to Datadog, the array is JSON-encoded [here](https://github.com/DataDog/opentelemetry-mapping-go/blob/b22e5064e03a26bb7b9056888ef69c90370abe65/pkg/otlp/metrics/dimensions.go#L84), then the tag rules in the [Datadog documentation](https://docs.datadoghq.com/developers/guide/what-best-practices-are-recommended-for-naming-metrics-and-tags/#rules-and-best-practices-for-naming-tags) are applied.

Thus, an array-valued attribute like `key=["value1", "value2"]` gets turned into the tag `key:_value1_value2`, which is unintuitive and difficult to filter on. For some use cases (including mine), mapping array attributes into a tag for each element, such as `key:value1` and `key:value2`, would be ideal and allow for more natural filtering within Datadog.

Could an option for this be added?

mx-psi:
@squah-confluent Thanks for your issue! The current behavior is to map the value of attributes using the `AsString()` function, that gives a JSON-like serialization of attributes. Datadog is restrictive as to what characters are supported in tags, so what should be `key:[value1,value2]` ends up as `key:_value1_value2` in Datadog.

Using the same key multiple times may lead to confusing behavior in the Datadog app. Additionally, in the OpenTelemetry world, this is not supported (the same key must not appear twice), so I would also not recommend this approach if you want to remain vendor agnostic.

Are you using the OpenTelemetry Collector Datadog exporter? Or are you using the Datadog Agent instead? I can help propose an alternative solution if you give more information about your setup.

squah-confluent:
@mx-psi Thanks, I understand using multiple tags with the same key isn't recommended.

I'm using the OpenTelemetry Collector with the Datadog exporter. I can't go into much detail about the metric itself. The tags come from another system which I have little control over and I'm trying to transfer them into Datadog faithfully. Most series have 0 or 1 instances of a tag and some have 2 or more. There's no ordering to the tag values and they're only intended to be used for filtering within Datadog. Some vendor lock-in is okay for this use case.

mx-psi:
@squah-confluent Thanks for the details! One more question: do you only need this for metrics? Or is there a need for this in other telemetry signals?

[...]

PR #331, opened 2024-05-17
[pkg/otlp/metrics] Add WithSplatArrayAttributes

mx-psi:
### What does this PR do?

Adds support for 'splatting' array attributes into multiple tags with the same key and different values

### Motivation

Updates #329


pkg/otlp/metrics/dimensions.go
```diff
@@ -78,11 +78,18 @@ func (d *Dimensions) OriginProductDetail() OriginProductDetail {
 }
 
 // getTags maps an attributeMap into a slice of Datadog tags
-func getTags(labels pcommon.Map) []string {
+func getTags(labels pcommon.Map, splat bool) []string {
 	tags := make([]string, 0, labels.Len())
 	labels.Range(func(key string, value pcommon.Value) bool {
-		v := value.AsString()
-		tags = append(tags, utils.FormatKeyValueTag(key, v))
+		if splat && value.Type() == pcommon.ValueTypeSlice {
```
mackjmr, line 84: Are there slices of `X` type that we should be handling differently/ testing for ?

mackjmr (approved):

pkg/otlp/metrics/dimensions.go
mx-psi: I don't think this will happen ever because of open-telemetry/opentelemetry-specification/pull/3858 but I implemented it just in case, nice suggestion :)

mx-psi:
Re-requesting review since the code changed a bit

PR #332, opened 2024-05-17
Add remaining entities to OriginIDFromAttributes

liustanley:
### What does this PR do?

Adds remaining entities from entity list: https://github.com/DataDog/datadog-agent/tree/main/comp/core/tagger#entity-ids.

I'm not sure if we have proper mappings for `container_image_metadata`, `ecs_task` and `host`.

### DIFF
--- a/pkg/otlp/attributes/attributes.go
+++ b/pkg/otlp/attributes/attributes.go
@@ -152,6 +152,7 @@ func TagsFromAttributes(attrs pcommon.Map) []string {
 func OriginIDFromAttributes(attrs pcommon.Map) (originID string) {
 	// originID is always empty. Container ID is preferred over Kubernetes pod UID.
 	// Prefixes come from pkg/util/kubernetes/kubelet and pkg/util/containers.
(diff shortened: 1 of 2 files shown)


pkg/otlp/attributes/attributes.go
```diff
@@ -154,8 +154,22 @@ func OriginIDFromAttributes(attrs pcommon.Map) (originID string) {
 	// Prefixes come from pkg/util/kubernetes/kubelet and pkg/util/containers.
```
mx-psi, line 154:
```suggestion
	// Prefixes come from pkg/util/kubernetes/kubelet and pkg/util/containers.
	// See https://github.com/DataDog/datadog-agent/tree/2b4f456/comp/core/tagger#entity-ids
```

mx-psi (commented):
I think we need container platform to review this. It looks sensible to me, but I don't know if the chosen attributes are the right one with the docs we have

dineshg13 (approved):

jackgopack4:
@liustanley is this still relevant?

PR #334, opened 2024-05-21
[pkg/otlp/attributes] Add Kubernetes DD tags to keep

liustanley:
### What does this PR do?

Adds Kubernetes DD tags to keep in TagsFromAttributes from https://docs.datadoghq.com/containers/kubernetes/tag/?tab=containerizedagent#out-of-the-box-tags


pkg/otlp/attributes/attributes.go
```diff
@@ -95,39 +95,77 @@ var (
 
 	// Kubernetes out of the box Datadog tags
 	// https://docs.datadoghq.com/containers/kubernetes/tag/?tab=containerizedagent#out-of-the-box-tags
+	// https://github.com/DataDog/datadog-agent/blob/d33d042d6786e8b85f72bb627fbf06ad8a658031/comp/core/tagger/taggerimpl/collectors/workloadmeta_extract.go
```
songy23, line 98: Can we expose those constants and use them here? Am concerned about the tags diverge moving forward

songy23: OK, probably worth leaving an issue or ticket to track it. Would be good to keep those constants in a single source of truth

songy23 (approved):

PR #340, opened 2024-06-14
Fix panics on invalid sized trace & span IDs

kevinnoel-be:
Fix for https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/33566

### DIFF
--- a/.chloggen/kevinnoel-be_fix-panic-on-invalid-trace-span-ids.yaml
+++ b/.chloggen/kevinnoel-be_fix-panic-on-invalid-trace-span-ids.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component (e.g. pkg/quantile)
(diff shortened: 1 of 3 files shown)


mx-psi (commented):
Thanks for the PR, it seems like there are some tests failing: https://github.com/DataDog/opentelemetry-mapping-go/actions/runs/9515937620/job/26238071289?pr=340#step:6:62 would you be able to take a look at those?

kevinnoel-be:
> Thanks for the PR, it seems like there are some tests failing: https://github.com/DataDog/opentelemetry-mapping-go/actions/runs/9515937620/job/26238071289?pr=340#step:6:62 would you be able to take a look at those?

Oops, sure I'll check that further on my side. Sorry for the noise 😅

kevinnoel-be:
Should be better now, this is how [some upstream process parses it](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/883801950f337e9157955c6a089e116b4ef5657f/pkg/ottl/contexts/internal/ids.go#L18).

mx-psi (approved):
Thanks!

PR #406, opened 2024-11-01
[OTEL-2182] Stop prefixing some HTTP metrics with otelcol

songy23:
### What does this PR do?

Stop prefixing `http_server_duration`, `http_server_request_size` and `http_server_response_size` with `otelcol_`

### Motivation

These metrics can be from SDKs rather than collector. Stop prefixing them to be consistent with https://opentelemetry.io/docs/collector/internal-telemetry/#lists-of-internal-metrics

### DIFF
--- a/.chloggen/metrics-prefix.yaml
+++ b/.chloggen/metrics-prefix.yaml
@@ -8,7 +8,7 @@ component: pkg/otlp/metrics
 note: Stop prefixing `http_server_duration`, `http_server_request_size` and `http_server_response_size` with `otelcol_`
 
 # The PR related to this change
-issues: []
+issues: [406]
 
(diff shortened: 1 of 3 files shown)


.chloggen/metrics-prefix.yaml
```diff
@@ -0,0 +1,16 @@
...
+change_type: breaking
+
+# The name of the component (e.g. pkg/quantile)
+component: pkg/otlp/metrics
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Stop prefixing `http_server_duration`, `http_server_request_size` and `http_server_response_size` with `otelcol_`
+
+# The PR related to this change
+issues: []
```
songy23, line 11:
```suggestion
issues: [406]
```

PR #408, opened 2024-11-21
test codecov integration

ankitpatel96:
try integrating codecov

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -72,10 +72,11 @@ jobs:
           files: testresults/
       - name: Upload coverage reports to Codecov
         uses: codecov/codecov-action@v4
+        if: always() && startsWith( matrix.go-version, '1.21' )
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
       - name: cleanup
-        run: rm -rf testresults/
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yaml
```diff
@@ -70,8 +70,14 @@ jobs:
...
           files: testresults/
+      - name: Upload coverage reports to Codecov
+        uses: codecov/codecov-action@b9fd7d16f6d7d1b5d2bec1a2887e65ceed900238 # 4.6.0
+        if: startsWith( matrix.go-version, '1.21' )
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
       - name: cleanup
-        run: rm -rf testresults/
+        if: always()
+        run: rm -rf testresults/ && find . -type f -name "*.cover" -exec rm -f {} +
```
jade-guiton-dd, line 80: Nit: I don't think you need a `-f` option here; the arguments are guaranteed to exist since they were just found by `-f`.

Also, out of curiosity, what is the reason for cleaning up files at the end of the job? Is there a risk of a job with files remaining from the previous one and messing things up?

jade-guiton-dd (approved):
LGTM. Just a nitpick and a question:

PR #496, opened 2025-01-29
Prerelease pkgs v0.25.0

ogaca-dd:
### What does this PR do?

Update package to v0.25.0

### DIFF
--- a/pkg/inframetadata/go.mod
+++ b/pkg/inframetadata/go.mod
@@ -3,7 +3,7 @@ module github.com/DataDog/opentelemetry-mapping-go/pkg/inframetadata
 go 1.22.0
 
 require (
-	github.com/DataDog/opentelemetry-mapping-go/pkg/otlp/attributes v0.24.0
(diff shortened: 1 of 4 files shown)


.chloggen/olivierg_gateway-usage-gauge.yaml
```diff
@@ -1,16 +0,0 @@
-# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
```
truthbk, line 1: why are we removing these?

mackjmr (approved):

truthbk (approved):
Looks good to me 👍  Just a question to make sure the change log entries getting removed is intentional.

.chloggen/olivierg_gateway-usage-gauge.yaml
ogaca-dd: `chloggen update -v v0.25.0` removed these files automatically so this is expected.

PR #516, opened 2025-02-12
Add HostFromAttributesHandler in MapLog

ogaca-dd:

### DIFF
--- a/pkg/otlp/logs/translator.go
+++ b/pkg/otlp/logs/translator.go
@@ -47,8 +47,8 @@ func NewTranslator(set component.TelemetrySettings, attributesTranslator *attrib
 	}, nil
 }
 
-func (t *Translator) hostNameAndServiceNameFromResource(ctx context.Context, res pcommon.Resource) (host string, service string) {
-	if src, ok := t.attributesTranslator.ResourceToSource(ctx, res, signalTypeSet, nil); ok && src.Kind == source.HostnameKind {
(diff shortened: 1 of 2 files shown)


pkg/otlp/logs/translator.go
```diff
@@ -65,14 +65,14 @@ func (t *Translator) hostFromAttributes(ctx context.Context, attrs pcommon.Map)
 }
 
 // MapLogs from OTLP format to Datadog format.
-func (t *Translator) MapLogs(ctx context.Context, ld plog.Logs) []datadogV2.HTTPLogItem {
+func (t *Translator) MapLogs(ctx context.Context, ld plog.Logs, hostFromAttributesHandler attributes.HostFromAttributesHandler) []datadogV2.HTTPLogItem {
```
songy23, line 68: This will lead to another circular dependency problem between Agent and collector. Does it make sense to add a new func like `MapLogsWithHost` while leaving `MapLogs` as-is?

ogaca-dd: There is no circular dependency (at least I didn't spot any), but indeed I have to update `datadog-agent` (PR is nearly ready) and `opentelemetry-collector-contrib` (PR is ready and is a one line change).
Let me know if you still think it is better to add `MapLogsWithHost` and remove it latter to reduce friction.

songy23: Got it, for this one it's less of a problem, you might still need to do update version in Agent -> merge PR -> use a commit hash version of Agent in Collector.

songy23 (approved):

PR #598, opened 2025-04-07
[OTEL-2506] Convert OTel instrumentation scope attributes to log attributes / metric tags

jade-guiton-dd:
### What does this PR do?

Adds support for OTel instrumentation scope attributes in `pkg/otlp/logs` and `pkg/otlp/metrics`, by converting them to log attributes and metric tags. In the latter case, this is locked behind the `WithInstrumentationScopeMetadataAsTags` option, which corresponds to the `instrumentation_scope_metadata_as_tags` setting in the Datadog exporter (but currently nothing in the serializerexporter, I will fix that in a separate PR).

### Motivation

Since last week's release, the OTel Collector has a feature gate which enables injecting scope attributes in its internal telemetry to clearly identify which Collector component emitted it. This makes it important for Datadog to be able to read and display these attributes to the user.

### Describe how you validated your changes

I added test cases to both parts of the code to validate the basic translation behavior.

### DIFF
--- a/.chloggen/jade-guiton-dd_expose-scope-attributes.yaml
+++ b/.chloggen/jade-guiton-dd_expose-scope-attributes.yaml
@@ -1,0 +1,17 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'enhancement'
+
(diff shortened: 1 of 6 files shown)


songy23 (approved):
LGTM

PR #635, opened 2025-05-14
[pkg/otlp/metrics] Correctly treat Summary counts as cumulative monotonic sums

mx-psi:


pkg/otlp/metrics/metrics_translator.go
```diff
@@ -578,11 +578,20 @@ func (t *Translator) mapSummaryMetrics(
...
 
-		// count and sum are increasing; we treat them as cumulative monotonic sums.
+		// treat count as a cumulative monotonic metric
+		// and sum as a non-monotonic metric
+		// https://prometheus.io/docs/practices/histograms/#count-and-sum-of-observations
 		{
 			countDims := pointDims.WithSuffix("count")
-			if dx, ok := t.prevPts.Diff(countDims, startTs, ts, float64(p.Count())); ok && !t.isSkippable(countDims.name, dx) {
-				consumer.ConsumeTimeSeries(ctx, countDims, Count, ts, dx)
+			val := float64(p.Count())
```
jackgopack4, line 586: I understand that TimeSeriesConsumer expects a float, but wondering if we should consider having an implementation that allows for int for count. Any negative side effects of casting uint to float? Could we have inaccurate counts at high values?

pkg/otlp/metrics/testdata/datadogdata/summary/simple_summary-with-quantile.json
```diff
@@ -82,4 +82,4 @@
       "Value": 600
     }
   ]
-}
+}
```
jackgopack4, line 85:
```suggestion
}

```

pkg/otlp/metrics/testdata/datadogdata/summary/simple_summary_cumsum-keep.json
```diff
@@ -0,0 +1,41 @@
...
+      "OriginID": "",
+      "OriginProduct": 10,
+      "OriginSubProduct": 17,
+      "OriginProductDetail": 0,
+      "Type": "count",
+      "Timestamp": 1669802350921840070,
+      "Value": 100
+    }
+  ]
+}
```
jackgopack4, line 41:
```suggestion
}

```

jackgopack4 (commented):
looks good just have a question about TimeSeries consumer and small tweak suggested for the json

pkg/otlp/metrics/metrics_translator.go
mx-psi: Datadog data model, at least at the intake API level, is based on floats (sent you an internal link with docs), so while I agree that this could lead to issues if the integer is big enough there's not much we can do about it without changes to the intake

PR #652, opened 2025-05-21
Send hostmap on new host

mackjmr:
### What does this PR do?

This fixes a bug when host meta was not sent for new hosts, which means if the host never changed, the host meta would never be sent.


mx-psi (approved):
Sounds reasonable but the tests are failing

PR #654, opened 2025-05-23
[OTEL-2649] Fix cumulative-to-delta conversion to avoid spikes on translator restart

jade-guiton-dd:
### Context

Because Datadog does not currently have a concept of cumulative count metrics, we perform cumulative-to-delta conversion in the metrics translator when consuming a monotonic cumulative sum metric. (In the non-monotonic cumulative case, we output a gauge instead). This involves storing the latest value of each metric series, and when receiving a new point, emitting the difference with the previous value.

This however begs the question of how to deal with the first point we receive from a given metric series, since in that scenario we don't have a previous value to compare against. If this is the very first point of the series, then the previous value is implicitly zero, and we can pass through the value as a valid delta. If it isn't, because the translator did not receive previous points from this series, then we can't compute a meaningful delta and must emit nothing.

The heuristic that was chosen to determine this is implemented in the `shouldConsumeInitialValue` function in `pkg/otlp/metrics/metrics_translator.go`: in the default "auto" mode, we emit the initial point as a delta if the start timestamp of the metric series is set, and comes _after_ the start of the metrics translator code.

(This assumes the main reason the translator may not receive points from a series is that it hadn't started yet. We also have options to always emit, or always drop the initial point to hopefully cover other cases.)

### What does this PR do?

This PR fixes the timestamp comparison in `shouldConsumeInitialValue`. Previously, `startTs` was compared to `getProcessStartTime()`, even though the former is in nanoseconds and the latter in seconds. This meant that in the most common scenario, we always treated the first received point as a valid delta, resulting in large "spikes" when the Agent/Collector containing the translator restarted but the application did not.

This PR also adds a new test which checks the correct behavior of `mapNumberMonotonicMetrics` for all three values of `InitialCumulMonoValueMode`, and in the cases of both an "app restart" (ie. the metric series started after the translator did), and an "agent restart" (ie. the metric series started before the translator did).

### DIFF
--- a/.chloggen/jade-guiton-dd_fix-cumul-to-delta-initial-value.yaml
+++ b/.chloggen/jade-guiton-dd_fix-cumul-to-delta-initial-value.yaml
@@ -1,0 +1,17 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'bug_fix'
+
+# The name of the component (e.g. pkg/quantile)
+component: pkg/otlp/metrics
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Fix automatic intial point dropping when converting cumulative monotonic sum metrics
+
+# The PR related to this change
+issues: [654]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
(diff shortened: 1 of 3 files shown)


songy23 (commented):
please add a changelog

songy23 (approved):
Need to fix linter too, otherwise LGTM

PR #661, opened 2025-05-27
Add support for host aliases

mackjmr:
### What does this PR do?

This PR adds support for host aliases. Users can now set host aliases by setting resource attribute `datadog.host.use_as_metadata` to true and `datadog.host.aliases` as a slice of host aliases.

### DIFF
--- a/pkg/inframetadata/internal/hostmap/tags.go
+++ b/pkg/inframetadata/internal/hostmap/tags.go
@@ -15,7 +15,10 @@ import (
 	conventions "go.opentelemetry.io/collector/semconv/v1.21.0"
 )
 
-const hostTagPrefix = "datadog.host.tag."
+const (
+	hostTagPrefix      = "datadog.host.tag."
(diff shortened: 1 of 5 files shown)


pkg/inframetadata/internal/hostmap/tags.go
```diff
@@ -63,3 +66,21 @@ func getHostTags(m pcommon.Map) ([]string, error) {
 	sort.Strings(tags)
 	return tags, err
 }
+
+// getHostAliases tries to get a host aliases from attribute datadog.host.aliases
+func getHostAliases(attrs pcommon.Map) []string {
+	var hostAliases []string
+	attrs.Range(func(k string, v pcommon.Value) bool {
+		if k == hostAliasAttribute {
+			hostAliasesAny := v.Slice().AsRaw()
```
songy23, line 75: nit: is this safe? Do you need a type assertion here?

PR #724, opened 2025-07-10
[chore] Add tests about which hosts are consumed

mx-psi:
### What does this PR do?

Add tests regarding which hosts are consumed.

### Motivation

Follow up to #721


jackgopack4:
will merge after CI passes again

PR #725, opened 2025-07-11
[breaking][pkg/otlp/metrics] Pass interval as part of ConsumeTimeSeries and ConsumeSketch function

mx-psi:
### What does this PR do?

Explicitly pass `interval` when consuming sketches and timeseries.

The interval type varies between implementations:
- [Zorkian uses `int`](https://pkg.go.dev/gopkg.in/zorkian/go-datadog-api.v2#Metric)
- [datadog-api-client uses `*int64`](https://pkg.go.dev/github.com/DataDog/datadog-api-client-go/v2@v2.19.0/api/datadogV2#MetricSeries)
- [serializer implementation uses `int64`](https://github.com/DataDog/datadog-agent/blob/66a632665c776c6f92397eb4e88f74dc88e143ba/pkg/metrics/series.go#L56)

I have decided to favor the serializer one (`int64`). [`0` is the same as unset](https://github.com/DataDog/datadog-agent/blob/66a632665c776c6f92397eb4e88f74dc88e143ba/comp/otelcol/otlp/components/exporter/serializerexporter/consumer.go#L146).

### Motivation

This allows us to pass interval data if we ever want to infer it or add it via attributes

### DIFF
--- a/.chloggen/mx-psi_delta-interval-infer.yaml
+++ b/.chloggen/mx-psi_delta-interval-infer.yaml
@@ -1,0 +1,17 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component (e.g. pkg/quantile)
+component: pkg/otlp/metrics
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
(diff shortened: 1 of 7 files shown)


mx-psi:
If the breakage is too big, I can instead add a new method and deprecate the old one

jade-guiton-dd (approved):

PR #744, opened 2025-08-04
Error on non monotonic bounds

mackjmr:
### What does this PR do?

Return error on non monotonic bounds and add test case.

### Motivation

OTAGENT-449


.chloggen/mackjmr_err-non-monotonic-bounds-deprecate.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: deprecation
+
+# The name of the component (e.g. pkg/quantile)
+component: pkg/quantile
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Deprecate InsertInterpolate in favor of InsertIntrp which returns an error on non monotonic bounds.
```
songy23, line 8: Can InsertInterpolate simply return an error instead of duplicating the funcs ? I don't think it's a breaking change to return an error there

mackjmr: good call, updated this 👍

PR #745, opened 2025-08-04
add OTLP -> RUM translation + RUM handling in logs exporter

dd-jasminesun:
### What does this PR do?
add OTLP signal attribute -> RUM event JSON translation to pkg and also update MapLogs so that the log exporter has RUM event handling 

### Motivation
https://datadoghq.atlassian.net/browse/OTEL-2690


pkg/otlp/logs/translator.go
```diff
@@ -79,19 +86,84 @@ func (t *Translator) MapLogs(ctx context.Context, ld plog.Logs, hostFromAttribut
 			scope := sl.Scope()
 			// iterate over Logs
 			for k := 0; k < lsl.Len(); k++ {
-				log := lsl.At(k)
+				logRecord := lsl.At(k)
+				//TODO (OTEL-2731): get forward_otlp_rum_to_dd_rum from config instead of setting to true
+				forward_otlp_rum_to_dd_rum := true
```
IbraheemA, line 91: Upon second look, this would have the side effect of changing the UX for existing users who are sending logs with `session.id`. (Possibly throwing concerning errors like "failed to create request"). Lets copy this entire `MapLogs` function and make a second function `MapLogsAndRouteRUMEvents`, which has an additional parameter `shouldForwardOTLPRUMToDDRUM`. We'll use that one in the logs_exporter. Add a comment to MapLogs that marks it as deprecated.

(We need to do it this way instead of changing the signature of MapLogs because we can't make breaking changes to the API).

pkg/otlp/rum/mapping.go
```diff
@@ -0,0 +1,307 @@
+package rum
+
+var OTLPToRUMAttributeMap = map[string]string{
```
dineshg13, line 3: nit: this look like DD to OTel mapping

pkg/otlp/rum/mapping.go
```diff
@@ -0,0 +1,307 @@
...
+	Datadog + "." + View + "." + CPUTicksCount:                                               View + "." + CPUTicksCount,
+	Datadog + "." + View + "." + CPUTicksPerSecond:                                           View + "." + CPUTicksPerSecond,
+	Datadog + "." + View + "." + RefreshRateAverage:                                          View + "." + RefreshRateAverage,
+	Datadog + "." + View + "." + RefreshRateMin:                                              View + "." + RefreshRateMin,
+	Datadog + "." + View + "." + SlowFramesRate:                                              View + "." + SlowFramesRate,
+	Datadog + "." + View + "." + FreezeRate:                                                  View + "." + FreezeRate,
+	Datadog + "." + View + "." + FlutterBuildTime + "." + Min:                                View + "." + FlutterBuildTime + "." + Min,
+	Datadog + "." + View + "." + FlutterBuildTime + "." + Max:                                View + "." + FlutterBuildTime + "." + Max,
+	Datadog + "." + View + "." + FlutterBuildTime + "." + Average:                            View + "." + FlutterBuildTime + "." + Average,
```
dineshg13, line 250: Am i reading this rt - the only diff is in prefix ? if so can we just add/remove prefix rather than this map

pkg/otlp/rum/rum.go
```diff
@@ -0,0 +1,66 @@
...
+		} else {
+			existing, ok := current[part]
+			if !ok {
+				current[part] = make(map[string]any)
+			} else if _, isMap := existing.(map[string]any); !isMap {
+				// force override if it's not a map
+				current[part] = make(map[string]any)
+			}
+			current = current[part].(map[string]any)
+		}
```
dineshg13, line 51: Avoid arrow code pattern. No else please . https://medium.com/@dfs.techblog/cleaner-code-tackling-arrowhead-anti-pattern-238d5ce91390

pkg/otlp/rum/rum.go
```diff
@@ -0,0 +1,66 @@
...
+
+func ConstructRumPayloadFromOTLP(attr pcommon.Map) map[string]any {
+	rumPayload := make(map[string]any)
+	attr.Range(func(k string, v pcommon.Value) bool {
+		if rumAttributeName, exists := OTLPToRUMAttributeMap[k]; exists {
+			buildRumPayload(rumAttributeName, v, rumPayload)
+		} else {
+			buildRumPayload(strings.TrimPrefix(k, "datadog."), v, rumPayload)
+		}
+		return true
```
dineshg13, line 63: Same, no else

pkg/otlp/rum/rum_test.go
```diff
@@ -0,0 +1,322 @@
...
+				"score": 3.14,
+			},
+		},
+		{
+			name: "bool value",
+			key:  "enabled",
+			value: func() pcommon.Value {
+				v := pcommon.NewValueBool(true)
+				return v
+			}(),
```
dineshg13, line 64: can't u do `value: pcommon.NewValueBool(true)` ?

pkg/otlp/rum/mapping.go
dd-jasminesun: it is from OTel to DD since attributes that have not been defined yet in the OTel spec will have the `datadog` prefix. ref: https://datadoghq.atlassian.net/wiki/spaces/OTEL/pages/5194580037/RUM+Payload+Attributes+-+OTLP+Signal+Attributes

.chloggen/rum.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: new_component
+
+# The name of the component (e.g. pkg/quantile)
+component: pkg/otlp/rum
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add OTLP -> RUM translation and also enhance MapLogs so that it can be used for RUM event handling
```
IbraheemA, line 8: Update now that we've made a new method. Mention that MapLogs is deprecated.

pkg/otlp/logs/transform_test.go
```diff
@@ -31,56 +31,28 @@ import (
...
+	return []translatorTestCase{
 		{
-			// different attribute types
 			name: "basic",
-			args: args{
+			args: struct {
+				lr    plog.LogRecord
+				res   pcommon.Resource
+				scope pcommon.InstrumentationScope
+			}{
```
IbraheemA: I would declare the "args" struct type at top-level so you don't have to redeclare it for every test case

pkg/otlp/logs/transform_test.go
```diff
@@ -673,15 +722,53 @@ func TestTranslator(t *testing.T) {
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			logs := plog.NewLogs()
+			rl := logs.ResourceLogs().AppendEmpty()
+			tt.args.res.MoveTo(rl.Resource())
+			sl := rl.ScopeLogs().AppendEmpty()
+			tt.args.scope.MoveTo(sl.Scope())
+			tt.args.lr.CopyTo(sl.LogRecords().AppendEmpty())
+
+			payloads := translator.MapLogsAndRouteRUMEvents(context.Background(), logs, nil, true)
```
IbraheemA, line 730: Also test with shouldForwardOTLPRUMToDDRUM=false

pkg/otlp/rum/mapping.go
```diff
@@ -0,0 +1,307 @@
...
+	Device + "." + Model + "." + Identifier:                                  Device + "." + Model,
+	Device + "." + Manufacturer:                                              Device + "." + Brand,
+	Datadog + "." + Device + "." + Architecture:                              Device + "." + Architecture,
+	Datadog + "." + Device + "." + Locale:                                    Device + "." + Locale,
+	Datadog + "." + Device + "." + Locales:                                   Device + "." + Locales,
+	Datadog + "." + Device + "." + TimeZone:                                  Device + "." + TimeZone,
+	Datadog + "." + Device + "." + BatteryLevel:                              Device + "." + BatteryLevel,
+	Datadog + "." + Device + "." + PowerSavingMode:                           Device + "." + PowerSavingMode,
+	Datadog + "." + Device + "." + BrightnessLevel:                           Device + "." + BrightnessLevel,
+	Datadog + "." + DD + "." + FormatVersion:                                 DD + "." + FormatVersion,
```
IbraheemA, line 52: Having two Datadog prefixes is redundant. Instead of `datadog._dd.format_version`, use just `datadog.format_version` for the key. Same for other mappings below.

IbraheemA (changes requested):

pkg/otlp/rum/mapping.go
dineshg13: This file scares me. it would be hard to maintain. If say for 90% of them it is prefix based, lets remove this from here do a prefix based mapping.

pkg/otlp/rum/mapping.go
dineshg13: This file scares me. it would be hard to maintain. If say for 90% of them it is prefix based, lets remove this from here do a prefix based mapping.

pkg/otlp/rum/rum.go
IbraheemA: @dd-jasminesun please use [switch statements](https://go.dev/wiki/Switch) instead of if/else-if

IbraheemA (approved):
LGTM! Let's wait for review from RUM team and then merge if all looks good

pkg/otlp/rum/mapping.go
```diff
@@ -0,0 +1,70 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2023-present Datadog, Inc.
+
+package rum
+
+var OTLPAttributeToRUMPayloadKeyMapping = map[string]string{
+	// _common-schema.json
```
BenoitZugmeyer, line 9: **suggestion:** add a comment to provide links to the [Datadog](https://github.com/DataDog/rum-events-format/tree/master/schemas) and OTLP formats reference

pkg/otlp/rum/mapping.go
```diff
@@ -0,0 +1,70 @@
...
+	Service + "." + Name:                                    Service,
```
BenoitZugmeyer, line 10: **question:** why using constants for every words instead of writing the field as a string directly? (i.e. `"service.name"`)
Binary size?

pkg/otlp/logs/translator.go
```diff
@@ -64,7 +95,198 @@ func (t *Translator) hostFromAttributes(ctx context.Context, attrs pcommon.Map)
...
+	parts = append(parts, ddEvpOriginParam.ParamKey+"="+getParamValue(rattrs, lattrs, ddEvpOriginParam))
+
+	ddRequestIdParam := ParamValue{ParamKey: "dd-request-id", SpanAttr: "", Fallback: randomID()}
+	parts = append(parts, ddRequestIdParam.ParamKey+"="+getParamValue(rattrs, lattrs, ddRequestIdParam))
+
+	return "/api/v2/rum?" + strings.Join(parts, "&")
+}
+
+// MapLogsAndRouteRUMEvents from OTLP format to Datadog format if shouldForwardOTLPRUMToDDRUM is true.
+func (t *Translator) MapLogsAndRouteRUMEvents(ctx context.Context, ld plog.Logs, hostFromAttributesHandler attributes.HostFromAttributesHandler, shouldForwardOTLPRUMToDDRUM bool, rumIntakeUrl string) ([]datadogV2.HTTPLogItem, error) {
```
BenoitZugmeyer, line 189: **question:** I'm not sure to understand where this function is used?

pkg/otlp/logs/translator.go
```diff
@@ -64,7 +95,198 @@ func (t *Translator) hostFromAttributes(ctx context.Context, attrs pcommon.Map)
...
+
+func randomID() string {
+	b := make([]byte, 16)
+	if _, err := rand.Read(b); err != nil {
+		panic(err)
+	}
+	return hex.EncodeToString(b)
+}
+
+func buildDDForwardURL(rattrs pcommon.Map, lattrs pcommon.Map) string {
```
BenoitZugmeyer, line 168: **suggestion:** don't call this `ddforward` as it has a special meaning (for proxying RUM data). Maybe `buildIntakeUrlPathAndParameters`

pkg/otlp/rum/mapping.go
```diff
@@ -0,0 +1,70 @@
...
+	User + "." + FullName:                                   Usr + "." + Name,
+	User + "." + Email:                                      Usr + "." + Email,
+	User + "." + Hash:                                       Usr + "." + AnonymousId,
+	User + "." + Name:                                       Account + "." + Name,
+	Os + "." + Name:                                         Os + "." + Name,
+	Os + "." + Version:                                      Os + "." + Version,
+	Os + "." + Build:                                        Os + "." + BuildId,
+	Device + "." + Model + "." + Name:                       Device + "." + Name,
+	Device + "." + Model + "." + Identifier:                 Device + "." + Model,
+	Device + "." + Manufacturer:                             Device + "." + Brand,
```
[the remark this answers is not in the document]
dd-jasminesun: is there a way that mobile users can send to a proxy like OTel collector? is that a common/used deployment pattern?

pkg/otlp/rum/mapping.go
```diff
@@ -0,0 +1,70 @@
...
+	Datadog + "." + BrowserSDKVersion:                             DD + "." + BrowserSDKVersion,
+	Datadog + "." + SDKName:                                       DD + "." + SDKName,
+
+	// action-schema.json
+	Datadog + "." + Action + "." + Position + "." + X:      DD + "." + Action + "." + Position + "." + X,
+	Datadog + "." + Action + "." + Position + "." + Y:      DD + "." + Action + "." + Position + "." + Y,
+	Datadog + "." + Action + "." + Target + "." + Selector: DD + "." + Action + "." + Target + "." + Selector,
+	Datadog + "." + Action + "." + Target + "." + Width:    DD + "." + Action + "." + Target + "." + Width,
+	Datadog + "." + Action + "." + Target + "." + Height:   DD + "." + Action + "." + Target + "." + Height,
+	Datadog + "." + Action + "." + NameSource:              DD + "." + Action + "." + NameSource,
```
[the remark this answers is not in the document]
dd-jasminesun: if a user is sending to otel collector with browser SDK, is it necessary for those internal fields to be set at export? if so, we would need to populate them somehow. currently we're passing the fields as attributes through the pipeline hence why we have the mappings.

pkg/otlp/logs/translator.go
```diff
@@ -64,7 +95,198 @@ func (t *Translator) hostFromAttributes(ctx context.Context, attrs pcommon.Map)
...
+	// sort the tags to ensure consistent ordering for testing purposes
+	sort.Strings(tagParts)
+
+	return strings.Join(tagParts, ",")
+}
+
+func randomID() string {
+	b := make([]byte, 16)
+	if _, err := rand.Read(b); err != nil {
+		panic(err)
```
IbraheemA, line 163: [panic should only be used for unrecoverable errors, it will end the process.](https://www.alexedwards.net/blog/when-is-it-ok-to-panic-in-go) Instead, return `(string, error)` and handle the error in the func outside

pkg/otlp/rum/mapping.go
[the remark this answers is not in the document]
IbraheemA: @BenoitZugmeyer  `DD RUM SDK -> collector -> DD backend` is probably not a useful pattern, but it may be useful to have these fields for `DD RUM SDK -> collector -> debug logs` or `(DD RUM SDK + OTel RUM SDK) -> collector -> DD backend`

PR #748, opened 2025-08-11
feat(otel-2740): deconstruct ddforward into resource

dd-jasminesun:
### What does this PR do?
As part of the effort to construct ddforward url in the exporter, we need to pass the ddforward url's query parameters as attributes in the RUM resource. The steps to construction are:
1. Deconstruct ddforward params into attributes in the receiver
2) In the exporter, construct ddforward; take whatever we can from OTel semantic conventions (e.g. `service`, `batch_time` (which might be able to use `date`), and `ddsource` /`dd-evp-origin` which can be taken from `source`); whatever is not available from semantic conventions, take from the deconstructed ddforward params and then have a fallback

### Motivation
The motivation for constructing the ddforward url in the exporter vs grabbing the entire ddforward url from the request and then setting that as a resource attribute is that when we're taking RUM from OTel browser SDKs, where there's no ddforward, our exporter logic should still work. 
https://datadoghq.atlassian.net/browse/OTEL-2740

### DIFF
--- a/pkg/otlp/rum/rum.go
+++ b/pkg/otlp/rum/rum.go
@@ -8,11 +8,11 @@ package rum
 import (
 	"encoding/binary"
 	"fmt"
+	"net/url"
 	"strconv"
 	"strings"
 
 	"go.opentelemetry.io/collector/pdata/pcommon"
-	semconv "go.opentelemetry.io/collector/semconv/v1.5.0"
 )
 
 func buildRumPayload(k string, v pcommon.Value, rumPayload map[string]any) {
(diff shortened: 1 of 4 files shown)


pkg/otlp/rum/rum.go
```diff
@@ -107,9 +107,43 @@ func parseIDs(payload map[string]any) (pcommon.TraceID, pcommon.SpanID, error) {
 	return uInt64ToTraceID(0, traceID), uInt64ToSpanID(spanID), nil
 }
 
-func parseRUMRequestIntoResource(resource pcommon.Resource, ddforward string) {
-	resource.Attributes().PutStr(semconv.AttributeServiceName, "browser-rum-sdk")
-	resource.Attributes().PutStr("request_ddforward", ddforward)
+func parseDDForwardIntoResource(attributes pcommon.Map, ddforward string) {
+	u, err := url.Parse("https://browser-intake-datadoghq.com" + ddforward)
```
IbraheemA, line 111: Please make this URL configurable - probably do this parse outside this method using RumIntakeURL or similar from config, and have parseDDForwardIntoResource accept the query object. (alternatively, strip the prefix and pass the string of query parameters into this method, or any other way that works - I leave it to you as an exercise)

IbraheemA (changes requested):

PR #765, opened 2025-08-19
[OTEL-2697][pkg/otlp/metrics] Add WithInferDeltaInterval option

mx-psi:
### What does this PR do?

Adds option to infer interval field. This only works for delta sums that have a `startTimestamp` set. We set the interval if and only if it is 'close' (within the tolerance, fixed to 0.1 seconds) to a whole number of seconds (since the Datadog API only supports a whole number of seconds as an interval). The tolerance accounts for jitter. The solution should be robust against network errors.

### Motivation

While ultimately we can only set the interval for users in a subset of cases, we can do so in this case reasonably well

### DIFF
--- a/pkg/otlp/metrics/metrics_translator.go
+++ b/pkg/otlp/metrics/metrics_translator.go
@@ -40,6 +40,10 @@ import (
 const (
 	metricName             string = "metric name"
 	errNoBucketsNoSumCount string = "no buckets mode and no send count sum are incompatible"
+
+	// intervalTolerance is the tolerance for interval calculation in seconds
+	// We use 0.1 seconds as tolerance to allow for some jitter.
+	intervalTolerance float64 = 0.1
 )
 
 var (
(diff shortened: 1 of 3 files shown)


pkg/otlp/metrics/metrics_translator.go
```diff
@@ -62,6 +66,31 @@ var (
...
+// is close to a whole number of seconds (within intervalTolerance), otherwise returns 0.
+func inferDeltaInterval(startTimestamp, timestamp uint64) int64 {
+	if startTimestamp == 0 {
+		// We can't infer the interval without the startTimestamp
+		return 0
+	}
+
+	// Convert nanoseconds to seconds
+	deltaSeconds := float64(timestamp-startTimestamp) / 1e9
+	if deltaSeconds < 0 {
```
jade-guiton-dd, line 80: If `startTimestamp` is after `timestamp`, the subtraction will overflow, returning a very large positive interval, so this condition will never trigger. You probably want to convert `timestamp` and `startTimestamp` to floats before the subtraction.

I think the "malformed data" test case currently passes because the computed delta is `18446744063.709553`, which fails the round tolerance check, also returning zero. Maybe consider modifying that test case to use `timestamp = 0`, `startTimestamp = 710000000`, which after overflow results in a timestamp very close to an integer number of seconds.

May also be prudent to add a more general check for extremely large intervals, for cases where both timestamps are present, but one of them is very wrong, the difference happens to be close to an integer number of seconds. Not sure how happy the backend would be with receiving a 1-year long interval.

jade-guiton-dd (changes requested):

pkg/otlp/metrics/metrics_translator.go
jade-guiton-dd: > I mean this can happen also with small intervals if we are unlucky, the hope (or what my analysis intended to show) is that the probability of that happening should be extremely small.

What was extremely unlikely was getting a delta outside the tolerance interval assuming that the data was sent regularly. If one of the timestamps is just completely wrong (maybe it's in seconds instead of nanoseconds?), the probability of hitting an integer is just 1 in 10.

But I see your point about trusting the source until proven otherwise.

jade-guiton-dd (approved):
Small bug in the test (and weird behavior in the breaking changes check) aside, LGTM

pkg/otlp/metrics/metrics_translator_test.go
```diff
@@ -2345,3 +2345,55 @@ var statsPayloads = []*pb.ClientStatsPayload{
...
+			name:     "no starttimestamp",
+			startTs:  0,
+			ts:       11e9,
+			expected: 0,
+		},
+		{
+			name:     "malformed data",
+			startTs:  710000000,
+			ts:       0,
+			expected: 1,
```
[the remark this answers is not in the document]
mx-psi: oh, duh, sorry, I intentionally changed this and forgot to change it back

